PR #11, opened 2024-04-23
Schematic code restructuring

yarikoptic:


asmacdo:
I took the jist, but not the code :)

PR #14, opened 2024-04-25
Add ci

asmacdo:
important stuff
 - Adds codespell to kickoff CI
 - add smoketest.sh (hand test)
 - refactoring from review

misc
 - cleanup cruft
 - fix pre-commit


yarikoptic:
Ok, let's consider it a start of CI testing, although IMHO better was to code it as some basic pytest test and keep so those matrix runs thanks etc

Issue #3, opened 2024-04-18
smon "redesign"


yarikoptic:
https://github.com/brainlife/abcd-spec/blob/master/hooks/smon is a great start for a desired script but we need to

- [x] make it capable of simply starting the job instead of "attaching itself" to the same session, so make it into a "wrapper" which starts the command similarly to how was done in https://github.com/con/duct/blob/main/duct_time#L13
   - e.g. `duct ./abandoning_parent.sh 5 ./consume_mem.py 100 200` 
   - it should use os.setsid() to establish a new session id
   - should be more careful in assessing which process is itself -- just match by PID? https://github.com/brainlife/abcd-spec/blob/master/hooks/smon#L161
- [x] start monitoring in a separate thread, while main thread would be "busy" executing the process
~- [ ] add to the json line records what type of record it is, e.g, `"record_type": "system-summary"`, `"processes-sample"`, `"processes-summary"`~
   - [ ] `"processes-summary"` should just provide at the end the summary across all the processes on peak memory and consumed CPU resources (similar to what we got with https://github.com/con/duct/blob/main/duct_time ) + `exit_code` of the child process, `"wall_run_time_sec"`
- [x] make it configurable 
   - [x] should use `argparse` (or any other built-in), but if user does not specify that particular option on command line -- should take from a corresponding environment variable. May be it would be as easy as providing `default` value as the one which comes from environment, e.g. default for `os.getenv('DUCT_OUTPUT_PREFIX', ".duct/run-logs/{iso_date_ms}_{process_id}")` . Here the default would be a "python f-string" and process would format it with a dict of known variables like that.
   - [x] `--output-prefix` (`DUCT_OUTPUT_PREFIX` env var), defaulting to what we have in https://github.com/con/duct/blob/main/duct_time#L3C25-L3C40 - time based etc. Make it an f-string.
   - [x] `--sample-interval SECONDS`, `--record-interval SECONDS` -- currently `smon` seems to do sampling every 2 seconds for 30 times, and selects maximum utilization in terms of CPU or memory among those samples, and then spits out the record per process.  Here we want to express not in number of samples to aggregate but rather how frequently to aggregate. May be just add to the record `"number_of_sample_aggregated" : INT` 
   - [x] `--record-types all,system-summary,processes-samples,processes-summary`  (`DUCT_RECORD_TYPES`) -- ','-separated list of which record type(s) to bother collecting. By default -- `all`
[...]

PR #18, opened 2024-04-29
Handle the outputs logic, run monitoring in thread, and make record types configurable

asmacdo:
Adds:
 - --outputs
 - --capture-outputs
 - --record-types
 - Tests for process_outputs
 
Runs the monitoring loop in a thread, which should prevent monitoring from blocking the main process, but this is not tested.

### DIFF
--- a/src/duct.py
+++ b/src/duct.py
@@ -28,7 +28,6 @@ class Report:
         self.number = 0
         self.system_info = {}
 
-    # TODO property?
     def collect_environment(self):
         self.env = (
             {k: v for k, v in os.environ.items() if k.startswith(ENV_PREFIXES)},
@@ -171,43 +170,46 @@ def create_and_parse_args():
         type=str,
         default="all",
         choices=["all", "none", "stdout", "stderr"],
-        help="Record stdout, stderr, both, or neither to log files.",
+        help="Record stdout, stderr, all, or none to log files.",
     )
     parser.add_argument(
         "--outputs",
         type=str,
         default="all",
         choices=["all", "none", "stdout", "stderr"],
-        help="print stdout, stderr, both, or neither to stdout/stderr respectively.",
+        help="Print stdout, stderr, all, or none to stdout/stderr respectively.",
     )
     parser.add_argument(
         "--record-types",
         type=str,
         default="all",
         choices=["all", "system-summary", "processes-samples"],
+        help="Record system-summary, processes-samples, or all",
     )
     return parser.parse_args()
 
 
 class TeeStream:
+    """TeeStream simultaneously streams to standard output (stdout) and a specified file."""
+
     def __init__(self, file_path):
         self.file = open(file_path, "w")
         (
-            self.master_fd,
-            self.slave_fd,
+            self.listener_fd,
+            self.writer_fd,
         ) = os.openpty()  # Use pseudo-terminal to simulate terminal behavior
 
     def fileno(self):
         """Return the file descriptor to be used by subprocess as stdout/stderr."""
-        return self.slave_fd
+        return self.listener_fd
 
     def start(self):
-        """Start a thread to read from the master_fd and write to stdout and the file."""
+        """Start a thread to read from the main_fd and write to stdout and the file."""
         thread = threading.Thread(target=self._redirect_output, daemon=True)
         thread.start()
 
     def _redirect_output(self):
-        with os.fdopen(self.master_fd, "rb", buffering=0) as stream:
+        with os.fdopen(self.listener_fd, "rb", buffering=0) as stream:
             while True:
                 try:
                     data = stream.read(1024)  # Read larger blocks of data
@@ -224,7 +226,7 @@ class TeeStream:
 
     def close(self):
         """Close the slave fd and the file when done."""
-        os.close(self.slave_fd)
+        os.close(self.listener_fd)
         self.file.close()
 
 
@@ -260,7 +262,7 @@ def main():
     os.makedirs(args.output_prefix, exist_ok=True)
 
     if args.capture_outputs in ["all", "stdout"] and args.outputs in ["all", "stdout"]:
-        stdout = TeeStream(args.output_prefix + "/stdout.txt")
+        stdout = TeeStream(f"{args.output_prefix}/stdout.txt")
         stdout.start()
     elif args.capture_outputs in ["none", "stderr"] and args.outputs in [
         "all",
@@ -271,7 +273,7 @@ def main():
         stdout = subprocess.DEVNULL
 
     if args.capture_outputs in ["all", "stderr"] and args.outputs in ["all", "stderr"]:
-        stderr = TeeStream(args.output_prefix + "/stderr.txt")
+        stderr = TeeStream(f"{args.output_prefix}/stderr.txt")
         stderr.start()
     elif args.capture_outputs in ["none", "stdout"] and args.outputs in [
         "all",
@@ -294,7 +296,6 @@ def main():
         report.get_system_info()
 
         if args.record_types in ["all", "processes-samples"]:
-            # TODO yoh is there a less ugly way to do this?
             monitoring_args = [
                 stdout,
                 stderr,
@@ -309,7 +310,7 @@ def main():
             )
             monitoring_thread.start()
             monitoring_thread.join()
-        #
+
         if args.record_types in ["all", "system-summary"]:
             with open(
                 f"{args.output_prefix}/system-report.session-{report.session_id}.json",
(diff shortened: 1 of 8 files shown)


asmacdo:
This PR should close out the last of the "smon redesign" checklist https://github.com/con/duct/issues/3

src/duct.py
```diff
@@ -164,62 +166,165 @@ def create_and_parse_args():
...
+        """Return the file descriptor to be used by subprocess as stdout/stderr."""
+        return self.listener_fd
+
+    def start(self):
+        """Start a thread to read from the main_fd and write to stdout and the file."""
+        thread = threading.Thread(target=self._redirect_output, daemon=True)
+        thread.start()
+
+    def _redirect_output(self):
+        with os.fdopen(self.listener_fd, "rb", buffering=0) as stream:
```
yarikoptic, line 213: is there any performance impact if there is heavy IO?  e.g. run with some dummy `cat largefile` and see CPU ... is it less of effect if buffering is enabled?

src/duct.py
```diff
@@ -164,62 +166,165 @@ def create_and_parse_args():
...
+                ) as resource_statistics_log:
+                    pinfo["elapsed_time"] = elapsed_time
+                    resource_statistics_log.write(json.dumps(aggregated))
+            report.number += 1
+        time.sleep(sample_interval)
+
+
+def prepare_outputs(capture_outputs, outputs, output_prefix):
+    if capture_outputs in ["all", "stdout"] and outputs in ["all", "stdout"]:
+        stdout = TeeStream(f"{output_prefix}/stdout.txt")
```
yarikoptic, line 264: let's not breed folders unless asked to:
- if `output_prefix` ends with `/` -- ensure that folder exists
- here/somewhere before if to do once do `output_prefix.format()` giving those variables to format
- just do `f"{output_prefix_formatted}stdout"` (and stderr below) so it would be for prefix to contain `_` separator. (I have reservations against .txt  since there is no guarantee that it is really a .txt -- may be it was some `dd` from a binary file, or may be it is json inside?... and probably no benefit from having extension)

src/duct.py
```diff
@@ -164,62 +166,165 @@ def create_and_parse_args():
...
+        stdout.start()
+    elif capture_outputs in ["none", "stderr"] and outputs in ["all", "stdout"]:
+        stdout = subprocess.PIPE
+    else:
+        stdout = subprocess.DEVNULL
+
+    if capture_outputs in ["all", "stderr"] and outputs in ["all", "stderr"]:
+        stderr = TeeStream(f"{output_prefix}/stderr.txt")
```
yarikoptic, line 272: same here -- no .txt and logic above for either a folder or not

src/duct.py
```diff
@@ -164,62 +166,165 @@ def create_and_parse_args():
...
-        process = subprocess.Popen(
-            [str(args.command)] + args.arguments.copy(),
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            preexec_fn=os.setsid,
+    stdout, stderr = prepare_outputs(
+        args.capture_outputs, args.outputs, args.output_prefix
+    )
+    process = subprocess.Popen(
+        [str(args.command)] + args.arguments.copy(),
```
yarikoptic, line 292: btw -- why need to copy?

src/duct.py
```diff
@@ -164,62 +166,165 @@ def create_and_parse_args():
...
-            if process.poll() is not None:  # the passthrough command has finished
-                break
-            time.sleep(args.sample_interval)
+        monitoring_thread.start()
+        monitoring_thread.join()
 
+    if args.record_types in ["all", "system-summary"]:
         with open(
-            f"{args.output_prefix}/system-report.session-{report.session_id}.json", "a"
+            f"{args.output_prefix}/system-report.session-{report.session_id}.json",
```
yarikoptic, line 320: following above -- might be not a directory, no need to make long filenames. best to keep session in common prefix unless you see multiple sessions for the same prefix, do you? if not do smth like

```suggestion
            f"{output_prefix_formatted}info.json",
```

src/duct.py
```diff
@@ -153,7 +155,7 @@ def create_and_parse_args():
         help="Interval in seconds between status checks of the running process.",
     )
     parser.add_argument(
-        "--output_prefix",
+        "--output-prefix",
         type=str,
         default=os.getenv("DUCT_OUTPUT_PREFIX", f".duct/run-logs/{file_safe_iso}"),
         help="Directory in which all logs will be saved.",
```
yarikoptic, line 161: - don't make it `f` right here -- keep it as a template
- extend doc to describe which variables are supported, e.g.  `file_safe_iso`, `session_id`
- make default to be `{file_safe_iso}-{session_id}_`

src/duct.py
```diff
@@ -164,62 +166,165 @@ def create_and_parse_args():
...
-
-    except Exception as e:
-        import traceback
-
-        traceback.print_exc()
-        import ipdb
-
-        ipdb.set_trace()
-        print(f"Failed to execute command: {str(e)}")
+    pprint.pprint(report, width=120)
```
yarikoptic, line 327: for now ok, but we might not want it to be reported... what would be informative -- point to the files produced (formatted prefix)

src/duct.py
asmacdo: Probably not necessary, but this way we don't have to trust that Popen wont modify

src/duct.py
asmacdo: > don't make it f right here -- keep it as a template

i don't follow rephrase?

yarikoptic: keep it as 

```
default=os.getenv("DUCT_OUTPUT_PREFIX", ".duct/run-logs/{file_safe_iso}")
```

so the value would be a template for `.format()` call so user could say e.g. `--output-prefix mylogs/{file_safe_iso}-{session_id}/`

src/duct.py
yarikoptic: it shouldn't ... IMHO too paranoid protection, adding then a question to the code reader either it was on purpose for some reason -- after all we do not do `.copy()` every time we pass a list into a function. So IMHO no reason to be special here unless there was a reason

src/duct.py
asmacdo: got it thanks

yarikoptic: I disagreed already -- I think we need to give flexibility to the user.
IMHO there should be little of complexity whatsoever.  As soon as you know session (without session you do not start anything) -- you are ready to format the path, so just do it and use that formatted `output_prefix_formated` from that point on. What complexity do you have in mind?

asmacdo: TeeStream needs to calculate the file path it will be writing to, and open the file prior to executing Popen.

src/duct.py
asmacdo: Ill resolve for now, we can open a new comment based on the changes.

src/duct.py
yarikoptic: so? decide then on the path before creating TeeStreams

asmacdo: WE cant do that if the path requires the session id.

asmacdo: We agreed not to support session_id here

src/duct.py
```diff
@@ -164,62 +166,165 @@ def create_and_parse_args():
...
 
+class TeeStream:
+    """TeeStream simultaneously streams to standard output (stdout) and a specified file."""
+
+    def __init__(self, file_path):
+        self.file = open(file_path, "w")
+        (
+            self.listener_fd,
+            self.writer_fd,
+        ) = os.openpty()  # Use pseudo-terminal to simulate terminal behavior
```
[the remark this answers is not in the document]
asmacdo: `$ ssh smaug ./duct-test.sh`

```
1
{"Command": "echo", "System": {"uid": "asmacdo", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
```

`$ ssh smaug cat -- duct-test.sh`

```
source ~/devel/duct/.duct-venv/bin/activate 
duct echo -- '1'
```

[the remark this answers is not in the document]
asmacdo: `ssh smaug "source ~/devel/duct/.duct-venv/bin/activate; duct echo -- '1'"     `
```
1
{"Command": "echo", "System": {"uid": "asmacdo", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
```

yarikoptic:
yeay -- I see some tests being added. Good!  Let's proceed so it is easier to review subsequent changes and try `duct`. So I will merge

PR #20, opened 2024-05-05
test for bug: TeeStream cannot handle high output (ie cat thousand_lines)

asmacdo:
First attempt at e2e testing of the TeeStream class. 

Currently not passing when using cat with more than 10^3 lines 

Theres a couple problems. [minor] First, it isn't deterministic on how much is written at a time, ie 250 lines one call, 300 the next, so asserting has_calls isn't working.

Bigger problem: somewhere between cat 100lines and cat 1000lines, TeeStream fails to print the whole thing before closing.

- Fixes Bug: Clean up fd closes


test/e2e/ten_1
```diff
@@ -0,0 +1,10 @@
+0
```
yarikoptic, line 1: FWIW -- please do not populate such simple to generate files within repo.  If needed -- a fixture with a session scope could create such files to be re-used across tests.

asmacdo:
Closing in favor of https://github.com/con/duct/pull/29

Issue #27, opened 2024-05-06
incomplete output of stdout is stored (when process died)


yarikoptic:
<details>
<summary>I had the process running which produced following output on the screen</summary> 

```shell
(git)smaug:~/proj/repronim/containers[master]git
$> datalad run -m "Update neurodesk image urls" duct scripts/replace_neurodesk_urls
[INFO   ] == Command start (output follows) =====
INFO: file images/neurodesk/neurodesk-afni--21.2.00.simg
OK: https://d15yxasja65rk8.cloudfront.net/afni_21.2.00_20210714.simg
 INFO: adding https://d15yxasja65rk8.cloudfront.net/afni_21.2.00_20210714.simg to images/neurodesk/neurodesk-afni--21.2.00.simg
^[[0Jaddurl https://d15yxasja65rk8.cloudfront.net/afni_21.2.00_20210714.simg ok
^[[1G INFO: removing 3 oracle urls
rmurl images/neurodesk/neurodesk-afni--21.2.00.simg ok
rmurl images/neurodesk/neurodesk-afni--21.2.00.simg ok
rmurl images/neurodesk/neurodesk-afni--21.2.00.simg ok
INFO: file images/neurodesk/neurodesk-afni--22.1.14.simg
ERROR: https://d15yxasja65rk8.cloudfront.net/afni_22.1.14_20220713.simg - could not verify presence: 403 . Will not be added
 INFO: removing 3 oracle urls
rmurl images/neurodesk/neurodesk-afni--22.1.14.simg ok
rmurl images/neurodesk/neurodesk-afni--22.1.14.simg ok
rmurl images/neurodesk/neurodesk-afni--22.1.14.simg ok
INFO: file images/neurodesk/neurodesk-afni--22.3.06.simg
 INFO: removing 3 oracle urls
ERROR: https://d15yxasja65rk8.cloudfront.net/afni_22.3.06_20221128.simg - could not verify presence: 403 . Will not be added
rmurl images/neurodesk/neurodesk-afni--22.3.06.simg ok
rmurl images/neurodesk/neurodesk-afni--22.3.06.simg ok
rmurl images/neurodesk/neurodesk-afni--22.3.06.simg ok
INFO: file images/neurodesk/neurodesk-afni--22.3.07.simg
 INFO: adding https://d15yxasja65rk8.cloudfront.net/afni_22.3.OK: https://d15yxasja65rk8.cloudfront.net/afni_22.3.07_20221206.simg
[...]

Issue #21, opened 2024-05-06
No point to print system information at the end -- print process aggregated one


yarikoptic:
relates also to
- #22 
At least the final record should know how long the process was running for! ATM there is no such information recorded -- only system one which is boring and not informative at all for the purpose of capturing the executed process stats.

PR #29, opened 2024-05-09
Refactor: Introducing... TailPipe!

### LINKED ISSUE #25: somehow adds 0d character into output dumps
- needs a test on some wild binary output ( may be just sweep through all 256 characters in order and after random permutation)  to correspond to original !!

I have noticed that after running 

```shell
$> duct -p duct-out_ -- bash -c "echo std output; echo 'err output' >&2"         
err output
std output
{"Command": "bash", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
```

I could not `grep .` those output files!!! 

```shell
$> grep . duct-out_*                              
duct-out_info.json:{"Command": "bash", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}{"Command": "bash", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}


```

although content is there... the culprit is that there is `0d` (13) is added for some reason!

```shell
$> hexdump -C duct-out_stdout                                             
00000000  73 74 64 20 6f 75 74 70  75 74 32 0d 0a           |std output2..|
0000000d
```

here is how file should look like (I just echo'ed into the file)

```shell
$> hexdump -C duct-out_stdout-manual 
00000000  73 74 64 20 6f 75 74 70  75 74 32 0a              |std output2.|
0000000c
```

### LINKED ISSUE #26: Do not obscure/change original stdout/stderr anyhow
Might be related to 
- #25

Should output **as is** it was received, no changes!  So if there are any ANSI terminal control characters (see https://en.wikipedia.org/wiki/ANSI_escape_code and #24) -- they should be passed as is.  I think the fact that output is somehow changed we get output from `git annex addurl` which likely uses some `\r` or may be some other ANSI command gets to look like

```shell
$> duct scripts/replace_neurodesk_urls
OK: https://d15yxasja65rk8.cloudfront.net/afni_21.2.00_20210714.simg
 INFO: adding https://d15yxasja65rk8.cloudfront.net/afni_21.2.00_20210714.simg to images/neurodesk/neurodesk-afni--21.2.00.simg
^[[0Jaddurl https://d15yxasja65rk8.cloudfront.net/afni_21.2.00_20210714.simg ok
^[[1G(recording state in git...)
 INFO: removing 108 oracle urls
rmurl images/neurodesk/neurodesk-afni--21.2.00.simg ok
rmurl images/neurodesk/neurodesk-afni--21.2.00.simg ok
rmurl images/neurodesk/neurodesk-afni--21.2.00.simg ok
{"Command": "scripts/replace_neurodesk_urls", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
```

Similarly to #25 the test could be to run duct running a command which outputs all kinds of "stuff" and ensuring that output we get to stdout from running duct is exactly what that command produces!

asmacdo:
TailPipe is nearly a drop-in replacement for TeeStream, but rather than "splitting the stream" we "tail -f" the stdout and stderr files as they are written.

BUG: should be streaming stdout and stderr, but instead we don't get output until the process ends.

Fixes: #27
Fixes #26
Fixes #25
Fixes https://github.com/con/duct/issues/21
Fixes https://github.com/con/duct/issues/10


asmacdo:
Fixes:
 - https://github.com/con/duct/issues/27
 - https://github.com/con/duct/issues/26
 - https://github.com/con/duct/issues/25
 - https://github.com/con/duct/issues/21

yarikoptic:
> BUG: should be streaming stdout and stderr, but instead we don't get output until the process ends.

It might be a show stopper to the approach.   Did you check if you could somehow tune up flushing or buffering behavior for those opened file(s) to ensure that stuff is dumped quickly after it is output by the process?  or could you just check if that is the case and may be the reading process just is not informed/can't read it for some time/some reason?

asmacdo:
> BUG: should be streaming stdout and stderr, but instead we don't get output until the process ends.

Fixed-- I was using write without flush to see what happens ;)

asmacdo:
Fixes: https://github.com/con/duct/issues/10

PR #33, opened 2024-05-27
Refactor start sampling

asmacdo:
I've corrected and cleaned up the sampling behavior. The most important difference in behavior that PID statistics will be appended to a file each <report_interval> (NDJSON). 

Now that we are producing better output, I also prettied it up a bit just for fun

image

### DIFF
--- a/src/duct.py
+++ b/src/duct.py
@@ -10,7 +10,7 @@ import subprocess
 import sys
 import threading
 import time
-from typing import Any, Dict, List, Optional, TextIO, Tuple, Union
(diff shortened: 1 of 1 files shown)


src/duct.py
```diff
@@ -97,57 +137,52 @@ def collect_sample(self):
                         "rss": int(rss),
                         # Virtual Memory size
                         "vsz": int(vsz),
+                        "timestamp": datetime.utcnow().isoformat() + "Z",
```
yarikoptic, line 140: let's use localized time stamps,
```suggestion
                        "timestamp": datetime.now().astimezone().isoformat(),
```

PR #35, opened 2024-05-30
Release from main not master here

yarikoptic:


yarikoptic:
codespell isn't happy -- I guess fixed in another Pr
```
Resulting CLI options  --skip ./.git
Error: ./src/duct.py:134: statestics ==> statistics
```

Issue #51, opened 2024-06-07
add [--] into --help output


yarikoptic:
so instead of
```
usage: duct [-h] [-p OUTPUT_PREFIX] [--sample-interval SAMPLE_INTERVAL] [--report-interval REPORT_INTERVAL] [-c {all,none,stdout,stderr}] [-o {all,none,stdout,stderr}]
            [-t {all,system-summary,processes-samples}]
            command [arguments ...]

```

looks like

```
usage: duct [-h] [-p OUTPUT_PREFIX] [--sample-interval SAMPLE_INTERVAL] [--report-interval REPORT_INTERVAL] [-c {all,none,stdout,stderr}] [-o {all,none,stdout,stderr}]
            [-t {all,system-summary,processes-samples}]
            [--] command [arguments ...]
```

and add description on `--` as to be recommended to ensure that `arguments` are not treated as arguments to `duct`.

On the other hand, I think we might need to look into how to tune argparser here: ideally we should automagically stop parsing when reaching `command`.  @jwodder , you did some heavy CLI parsing tune up in datalad-installer, may be you know what would be the best way here to ensure that parsing stops and `[arguments]` for the `command` do not interfer with our `duct ` arguments?

jwodder:
@yarikoptic Unless `duct` is calling [`parser.parse_intermixed_args()`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_intermixed_args), I think it should stop at the first non-option argument by default.

asmacdo:
duct is using  `parser.parse_args()`

https://github.com/con/duct/blob/0eec46d9c7527a7f17d3205447223f27b622aa87/src/duct.py#L271

yarikoptic:
not quite -- it does consume `[arguments]` and spits out error, e.g. we get

![image](https://github.com/con/duct/assets/39889/05e31be1-58b9-45eb-b8b6-4a1895347f51)

for 
```
yoh@typhon:/tmp/repro-TccSlQF/ds000003-qc$ cat .datalad/config 
[datalad "dataset"]
        id = 621f77c3-4a35-4147-8146-67fa49b66369
[datalad "containers.bids-mriqc"]
        image = containers/images/bids/bids-mriqc--0.16.0.sing
        cmdexec = duct {img_dspath}/containers/scripts/singularity_cmd run {img} {cmd}
```

in repronim/containers - I had to add `--` after duct there

jwodder:
[...]

PR #52, opened 2024-06-07
Pass unknown args to inner command

asmacdo:
Fixes: https://github.com/con/duct/issues/51

It works-- but I'm marking as a draft to investigate what happens if the inner command just happens to have the same named arg as `duct` does.


asmacdo:
I think this would break for inner commands that share arg names with `duct`. closing.

PR #56, opened 2024-06-07
Pass all args after <innercmd> to inner command

asmacdo:
Thanks for the idea @jwodder, works great :)

Fixes: https://github.com/con/duct/issues/51

### DIFF
--- a/src/duct.py
+++ b/src/duct.py
@@ -268,6 +268,7 @@ def create_and_parse_args():
         choices=["all", "system-summary", "processes-samples"],
         help="Record system-summary, processes-samples, or all",
     )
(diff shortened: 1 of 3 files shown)


src/duct.py
```diff
@@ -268,6 +268,7 @@ def create_and_parse_args():
         choices=["all", "system-summary", "processes-samples"],
         help="Record system-summary, processes-samples, or all",
     )
+    parser.add_argument("inner_args", nargs=argparse.REMAINDER)
```
jwodder, line 271: Shouldn't the `nargs` be applied to the `arguments` argument above instead of adding this?

asmacdo: Nice catch.

PR #75, opened 2024-06-10
Rewrite README update script in Python and remove from pre-commit

jwodder:
* `.update-readme-help.sh` benefits from being rewritten in Python because:
    * The syntax for `sed -i` with no extension differs between Linux and macOS, and thus the current shell script is not portable.
    * The Python script is more efficient, as it only needs to pass over the file once, where as the shell script processes the file twice.

* Running `.update-readme-help.sh` via pre-commit is problematic, as it requires `duct` to be installed in the current Python environment, yet no provision is made for this.  Unless you can come up with a simple, clean way to install `duct` in editable mode in some virtualenv, I think it's better to just leave this out of pre-commit.

### DIFF
--- a/.github/workflows/readme-updated.yaml
+++ b/.github/workflows/readme-updated.yaml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Update README.md helptext
         run: |
-          ./.update-readme-help.sh
+          ./.update-readme-help.py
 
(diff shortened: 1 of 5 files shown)


asmacdo:
Ah thank you, it was clunky and driving me crazy :)

PR #77, opened 2024-06-10
Move scripts used by tests to `test/data/`

jwodder:


asmacdo:
Thanks Ive been meaning to do this move. Also +1 the test_script docs are no longer needed.

PR #85, opened 2024-06-10
Autoupdate .pre-commit-config.yaml

jwodder:
At the very least, flake8 needs to be updated in order to properly parse certain f-strings in Python 3.12.


.pre-commit-config.yaml
```diff
@@ -4,30 +4,30 @@
 exclude: (profilers|sandbox|\.chatgpt)
```
asmacdo, line 4:
```suggestion
```

While we're in here lets remove these artifacts.

jwodder: Done: 59839f9

PR #91, opened 2024-06-10
Set `fail-fast` in `test.yml` back to `false`

jwodder:
`fail-fast` in `test.yml` was set to `true` in 213eabd in #14.  The commit message states that the change was temporary, yet this part was not reverted in bec8c95.  This PR sets `fail-fast` back to `false` so that PR authors will properly be informed if more than one test job fails.


jwodder:
@asmacdo Ping.  This would be really useful for debugging #90.

PR #102, opened 2024-06-11
Convert module to a package directory with a `py.typed` file

jwodder:
The module needs to be converted to a package directory so that it can contain a `py.typed` file, which is necessary so that mypy can type-check uses of `duct` items in the test code.


asmacdo (commented):
Looks reasonable to me, however @yarikoptic (at least when we first started) wanted the whole thing in a single python file that could be executed directly without even needing to install anything, including duct itself.

I guess we could move __version__ back to __main__? This looks more idiomatic though, @yarikoptic do you still want "executable as a single file script" to be a requirement?

yarikoptic:
So we might indeed refactor it to reside in multiple files but not sure if we are reaching threshold yet -- couldn't we substitute `__version__` within `__main__.py` directly (if you really want to be able to `python -m duct` it)

asmacdo:
I'd vote to go ahead and leave '__version__' in '__init__' so we don't build up weirdness to address later. If the no-install use case is requested by users maybe we revisit?

yarikoptic:
ok

Issue #23, opened 2024-05-06
Add option `--clobber` and by default error out if there is already report(s) at the location


yarikoptic:
ATM it causes inconsistent behavior overall by default - appends to `.json` (note and fix also -- no new line is recorded in that .json after final entry -- should be ended with a newline IMHO)  and overwrites outputs files

```
smaug:/tmp
$> rm duct-out*          
(dev3) 3 12170.....................................:Mon 06 May 2024 08:48:03 AM EDT:.
smaug:/tmp
$> duct -p duct-out_ -- bash -c "echo std output; echo 'err output' >&2"         
err output
std output
{"Command": "bash", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
(dev3) 3 12171.....................................:Mon 06 May 2024 08:48:17 AM EDT:.
smaug:/tmp
$> grep o duct-out_*
duct-out_info.json:{"Command": "bash", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
duct-out_stderr:err output
duct-out_stdout:std output
(dev3) 3 12172.....................................:Mon 06 May 2024 08:48:29 AM EDT:.
smaug:/tmp
$> duct -p duct-out_ -- bash -c "echo std output2; echo 'err output2' >&2"
err output2
std output2
{"Command": "bash", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
(dev3) 3 12173.....................................:Mon 06 May 2024 08:48:36 AM EDT:.
smaug:/tmp
$> grep o duct-out_*                                                      
duct-out_info.json:{"Command": "bash", "System": {"uid": "yoh", "memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}{"Command": "bash", "System": {"uid": "yoh", 
"memory_total": 135060111360, "cpu_total": 16}, "ENV": [{}], "GPU": []}
duct-out_stderr:err output2
duct-out_stdout:std output2
```

PR #103, opened 2024-06-12
Refactor File handling

asmacdo:
This PR centralizes all of the filepath formatting, checking for prior existence (adds clobber option), and creating new files.

This removes some complexity from many functions, and since it touchs a lot of the codebase the unittests required big changes.

Fixes: https://github.com/con/duct/issues/23


src/duct/__main__.py
```diff
@@ -472,13 +504,24 @@ def safe_close_files(file_list: Iterable[Any]) -> None:
             pass
 
 
-def ensure_directories(path: str) -> None:
+def ensure_directories(path: str, clobber: bool = False) -> None:
+    # Enforcing no path* files is perhaps overzealous, but this helps prevent
+    # conflicts between versions, prevents probable user errors, and leaves
+    # room for plugins down the road.
```
asmacdo, line 510: @yarikoptic feedback on possible overzealousness requested :)

yarikoptic: I would have not used glob here but "explicit better than implicit" operated only on the file names we care to create/write.  For that I would centralize generation of those filenames, smth like some `@dataclass` structure with `.stdout, .stderr, .info, .stats` attributes with those filenames filled-in/expanded based on the prefix. Then you could go through all of them and verify that none of those particular files exist. That would also centralize logic of minting those filenames which is now "append some suffix and the point of use".

asmacdo: Refactored out all the file handling, which touched nearly everything (and therefore proved that the old way was evil)

src/duct/__main__.py
```diff
@@ -13,7 +13,7 @@
 import sys
 import threading
 import time
-from typing import IO, Any, TextIO
+from typing import IO, Any, Generator, TextIO
```
jwodder, line 16: [`typing.Generator`](https://docs.python.org/3/library/typing.html#typing.Generator) is deprecated.  Use `collections.abc.Generator` instead.

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+    stderr: str
+    usage: str
+    info: str
+
+    def __iter__(self) -> Generator:
+        for each in fields(self):
+            yield each.name, getattr(self, each.name)
+
+    def __post_init__(self) -> None:
+        self._prefix = ""
```
jwodder, line 118: You should declare this as an attribute alongside `stdout` etc. with `_prefix: str = field(init=False)`.

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+
+    @property
+    def prefix(self) -> str:
+        return self._prefix  # type: ignore
```
jwodder, line 122: Declaring `_prefix` as an attribute should let you get rid of the comment here.

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
         )
 
 
+@dataclass
+class Suffix:
```
jwodder, line 95: If the only purpose of this class is to iterate over the same values every time, it'd be better to replace it with something like a constant list of `str` pairs (or, as I used in my suggestions for `LogPaths.create()`, a set of constants named `STDOUT_SUFFIX`, `STDERR_SUFFIX`, etc.).

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+
+
+@dataclass
+class LogPaths:
+    stdout: str
+    stderr: str
+    usage: str
+    info: str
+
+    def __iter__(self) -> Generator:
```
jwodder, line 113:
```suggestion
    def __iter__(self) -> Iterator[tuple[str, str]]:
```

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+        for each in fields(self):
```
jwodder, line 114: I think this would be cleaner to implement using `asdict()` instead of `fields()`.

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+
+    @classmethod
+    def create(cls, output_prefix: str, pid: None | int = None) -> LogPaths:
+        datetime_filesafe = datetime.now().strftime("%Y.%m.%dT%H.%M.%S")
+        formatted_prefix = output_prefix.format(
+            pid=pid, datetime_filesafe=datetime_filesafe
+        )
+        args_dict = {name: f"{formatted_prefix}{suffix}" for name, suffix in Suffix()}
+        new = cls(**args_dict)
+        new._prefix = formatted_prefix
```
jwodder, line 132: Why not just make `_prefix` a normal attribute of `LogPaths` that is initialized from a constructor argument instead of doing this stupid dance?  If your goal is just to keep `prefix` out of `self.__iter__()`, you can filter it out in that method.

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+        return self._prefix  # type: ignore
+
+    @classmethod
+    def create(cls, output_prefix: str, pid: None | int = None) -> LogPaths:
+        datetime_filesafe = datetime.now().strftime("%Y.%m.%dT%H.%M.%S")
+        formatted_prefix = output_prefix.format(
+            pid=pid, datetime_filesafe=datetime_filesafe
+        )
+        args_dict = {name: f"{formatted_prefix}{suffix}" for name, suffix in Suffix()}
+        new = cls(**args_dict)
```
jwodder, line 131: I would prefer explicitly spelling out each constructor argument separately like so:

```python
stdout = f"{formatted_prefix}{STDOUT_SUFFIX}"
stderr = f"{formatted_prefix}{STDERR_SUFFIX}"
# etc.
new = cls(stdout=stdout, stderr=stderr, ...)
```

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+        new._prefix = formatted_prefix
+        return new
+
+    def prepare_paths(self, clobber: bool, capture_outputs: Outputs) -> None:
+        conflicts = [path for path in asdict(self).values() if Path(path).exists()]
+        if conflicts and not clobber:
```
jwodder, line 137: I argue the logic here would be easier to follow if it were restructured as:

```python
if conflicts:
    if not clobber:
        raise FileExistsError(...)
    else:
        # Unlinking
```

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+        formatted_prefix = output_prefix.format(
+            pid=pid, datetime_filesafe=datetime_filesafe
+        )
+        args_dict = {name: f"{formatted_prefix}{suffix}" for name, suffix in Suffix()}
+        new = cls(**args_dict)
+        new._prefix = formatted_prefix
+        return new
+
+    def prepare_paths(self, clobber: bool, capture_outputs: Outputs) -> None:
+        conflicts = [path for path in asdict(self).values() if Path(path).exists()]
```
jwodder, line 136:
```suggestion
        conflicts = [path for _, path in self if os.path.exists(path)]
```

You have an `__iter__`, might as well use it.

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+        if conflicts and not clobber:
+            raise FileExistsError(
+                "Conflicting files:\n"
+                + "\n".join(f"- {path}" for path in conflicts)
+                + "\nUse --clobber to overrwrite conflicting files."
+            )
+        elif conflicts and clobber:
+            for path in conflicts:
+                Path(path).unlink()
```
jwodder, line 145:
```suggestion
                os.remove(path)
```

src/duct/__main__.py
```diff
@@ -91,6 +91,71 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+
+        if self.prefix.endswith(os.sep):  # If it ends in "/" (for linux) treat as a dir
+            os.makedirs(self.prefix, exist_ok=True)
+        else:
+            # Path does not end with a separator, treat the last part as a filename
+            directory = os.path.dirname(self.prefix)
+            if directory:
+                os.makedirs(directory, exist_ok=True)
+        for path_name, path in self:
+            if path_name in capture_outputs or capture_outputs == Outputs.ALL:
```
jwodder, line 155: I find the use of `path_name in capture_outputs` here rather obtuse.  It only works because `Outputs` inherits `str`, and the fact that `"stdout" not in Outputs.ALL` is counter-intuitive at first glance.  There's also the fact that, once you realize we're working with strings, `==` would make more sense than `in`.

I would suggest giving `Outputs` a `has(self, stream_name: str) -> bool` method that encapsulates this line's logic but compares `stream_name` against `self.value` using `==`.

src/duct/__main__.py
```diff
@@ -224,8 +290,8 @@ def collect_sample(self) -> Sample:
 
     def write_pid_samples(self) -> None:
         assert self._sample is not None
-        resource_stats_log_path = f"{self.output_prefix}usage.json"
-        with open(resource_stats_log_path, "a") as resource_statistics_log:
+
```
jwodder, line 293: Remove stray empty line.

src/duct/__main__.py
```diff
@@ -142,13 +207,14 @@ def __init__(
...
         self.system_info: SystemInfo | None = None
-        self.output_prefix = output_prefix
+        self.log_paths = log_paths
         self.max_values = Sample()
         self.process = process
         self._sample: Sample | None = None
-        self.datetime_filesafe = datetime_filesafe
         self.end_time: float | None = None
         self.run_time_seconds: str | None = None
+        self._usage_file_exists: bool = False
```
jwodder, line 216: This attribute is unused.

jwodder (changes requested):

src/duct/__main__.py
asmacdo: I didn't want _prefix to be treated as a `field`, so we could iterate over LogPaths and each would be a full path.

src/duct/__main__.py
asmacdo: Why is this preferred? (Unit tests will be easier, is that the main reason?) I went with Path().unlink since I am using Path().touch for consistency.

jwodder: If you're not already using `pathlib` for all your path values, there's not much point in converting string paths to a `Path` just for an operation that could be done just as well without the conversion.  (Also, I think `Path.touch()` might be replaceable with `os.utime()`; not entirely sure if it creates missing files.)

src/duct/__main__.py
asmacdo: as you said below, I'll just filter it out in __iter__

conftest.py
```diff
@@ -0,0 +1,10 @@
+import os
```
jwodder, line 1: This file should be under `test/`.

src/duct/__main__.py
asmacdo: As suggested below, we don't need Generator at all, Iterator is better.

src/duct/__main__.py
```diff
@@ -17,6 +16,17 @@
 from . import __version__
 
 ENV_PREFIXES = ("PBS_", "SLURM_", "OSG")
+STDOUT_SUFFIX = "stdout"
+STDERR_SUFFIX = "stderr"
+USAGE_SUFFIX = "usage.json"
+INFO_SUFFIX = "info.json"
+
+SUFFIXES = {
```
jwodder, line 24: An alternative I just thought of: Add a `LogFile` (or some other name) enum with `STDOUT`, `STDERR`, `USAGE`, and `INFO` variants plus a method for getting an instance's suffix.

src/duct/__main__.py
```diff
@@ -17,6 +16,17 @@
 from . import __version__
 
 ENV_PREFIXES = ("PBS_", "SLURM_", "OSG")
+STDOUT_SUFFIX = "stdout"
```
jwodder, line 19: You're not using these `SUFFIX` constants.  Get rid of them.

src/duct/__main__.py
asmacdo: especially now that we are filtering out prefix by key,  +1

src/duct/__main__.py
jwodder: This was not addressed.

src/duct/__main__.py
asmacdo: I was "looking ahead" imagining plugins that might want to create additionally files. But until then, lets keep it simple, +1

src/duct/__main__.py
jwodder: This was not addressed.

test/test_execution.py
```diff
@@ -26,11 +17,17 @@ def test_sanity_green(temp_output_dir: str) -> None:
         capture_outputs=Outputs.ALL,
         outputs=Outputs.ALL,
         record_types=RecordTypes.ALL,
+        clobber=False,
     )
     execute(args)
-    # When runtime < sample_interval, we won't have a usage.json
-    expected_files = ["stdout", "stderr", "info.json"]
+    expected_files = [
+        SUFFIXES["stdout"],
```
jwodder, line 24: Code like this would be a good place to use the `*_SUFFIX` constants, if you keep them.

jwodder (changes requested):

src/duct/__main__.py
asmacdo: I made this change, and then reverted. The surrounding logic is simpler now for other reasons, so I left this line as is. If it bothers you, I dont mind changing.

src/duct/__main__.py
asmacdo: Rather than use touch() I now use open(path, "w").close() which has the behavior I actually want: truncate if it exists, create it if it  doesnt. ie, Truncate replaces remove.

src/duct/__main__.py
asmacdo: How is that different than a dataclass with __iter__?

jwodder: Your `Suffix` dataclass only had once meaningful instance, which contained one field per logfile.  My suggestion is to create an enum with four instances, one instance per logfile.  This enum could then be used to replace the `stdout` etc. fields of `LogPaths` with a `paths: dict[LogFile, str]`, among other changes.

asmacdo: I decided to leave with the SUFFIX dict.

2 alt implementations from side chats (just for note taking)

```python
class LogSuffix(Enum):
    STDOUT = 1
    STDERR = 2
    USAGE = 3
    INFO = 4
 
    def suffix(self) -> str:
        if self is LogSuffix.STDOUT:
            return "stdout"
        elif # etc.
```

```python
class LogSuffix(Enum):
    STDOUT = "stdout"
    STDERR = "stderr"
    USAGE = "usage.json"
    INFO = "info.json"


@dataclass
class LogPaths:
    paths: dict[LogSuffix, str]
    prefix: str

    @classmethod
    def create(cls, output_prefix: str, pid: None | int = None) -> LogPaths:
        datetime_filesafe = datetime.now().strftime("%Y.%m.%dT%H.%M.%S")
        formatted_prefix = output_prefix.format(
            pid=pid, datetime_filesafe=datetime_filesafe
        )
        paths = {suffix: f"{formatted_prefix}{suffix.value}" for suffix in LogSuffix}
        return cls(paths, formatted_prefix)
```

jwodder (approved):

PR #112, opened 2024-07-09
Add pypi keywords

asmacdo:
- Add pypi keywords
- fixup .gitignore

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,15 +10,14 @@ license = MIT
 license_files = LICENSE
 url = https://github.com/con/duct/
 
-# TODO(asmacdo)
-# keywords =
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -10,15 +10,14 @@ license = MIT
...
-#     miniconda
-#     neurodebian
+keywords =
+    metrics
+    command-line
+    provenance
+    memory
+    cpu
+    usage
+    time
```
yarikoptic, line 20: I would sort and add `output-capture`

```suggestion
    command-line
    cpu
    memory
    metrics
    output-capture
    provenance
    time
    usage
```

PR #110, opened 2024-06-22
Improve usage.json schema

asmacdo:
Fixes https://github.com/con/duct/issues/109

```json
{
  "timestamp": "2024-06-22T12:05:35.337336-05:00",
  "num_samples": 1,
  "processes": {
    "1243562": {
      "pcpu": 0.0,
      "pmem": 0.0,
      "rss": 1760,
      "vsz": 221584,
      "timestamp": "2024-06-22T12:05:35.337336-05:00"
    }
  },
  "totals": {
    "pmem": 0.0,
    "pcpu": 0.0,
    "rss_kb": 1760,
    "vsz_kb": 221584
  },
  "averages": {
    "rss": 1760,
    "vsz": 221584,
    "pmem": 0.0,
    "pcpu": 0.0,
    "num_samples": 1
  }
}
{
  "timestamp": "2024-06-22T12:05:36.355308-05:00",
  "num_samples": 8,
  "processes": {
    "1243562": {
      "pcpu": 0.0,
      "pmem": 0.0,
      "rss": 1760,
      "vsz": 221584,
      "timestamp": "2024-06-22T12:05:36.355308-05:00"
    }
  },
  "totals": {
    "pmem": 0.0,
    "pcpu": 0.0,
    "rss_kb": 1760,
    "vsz_kb": 221584
  },
  "averages": {
    "rss": 1760.0,
    "vsz": 221584.0,
    "pmem": 0.0,
    "pcpu": 0.0,
    "num_samples": 8
  }
}

PR #113, opened 2024-07-25
Prepare for auto-powered releases

asmacdo:
I followed the instructions for a project that is already released: https://intuit.github.io/auto/docs/welcome/getting-started#make-latest-release

Now, it appears to be configured correctly. 

```
./auto info                

Environment Information:

"auto" version: v11.1.5
"git"  version: v2.45.0
"node" version: v16.16.0

Project Information:

✔ Repository:      con/duct
✔ Author Name:     Austin Macdonald
✔ Author Email:    austin@dartmouth.edu
✔ Current Version: v0.0.1
✔ Latest Release:  v0.0.1

✔ Labels configured on GitHub project 

GitHub Token Information:

✔ Token:            [Token starting with ghp_]
✔ Repo Permission:  admin
✔ User:             asmacdo
✔ API:              
✔ Enabled Scopes:   repo, write:packages
✔ Rate Limit:       2862/5000
```

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -13,4 +13,14 @@
             }
         ]
     ]
+    "labels": [
+      { "releaseType": "major", "name": "semver-major" },
+      { "releaseType": "minor", "name": "semver-minor" },
+      { "releaseType": "patch", "name": "semver-patch" },
+      { "releaseType": "none", "name": "semver-dependencies" },
+      { "releaseType": "none", "name": "semver-documentation" },
+      { "releaseType": "none", "name": "semver-internal" },
+      { "releaseType": "none", "name": "semver-performance" },
+      { "releaseType": "none", "name": "semver-tests" }
+    ]
 }
--- a/.autorc
+++ b/.autorc
@@ -1,6 +1,6 @@
 {
     "onlyPublishWithReleaseLabel": true,
-    "baseBranch": "master",
+    "baseBranch": "main",
     "author": "auto <auto@nil>",
     "noVersionPrefix": false,
     "plugins": [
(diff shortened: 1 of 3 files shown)


asmacdo:
I've also run `auto shipit --dry-run` and that looks good. I suppose we merge this, then I'll mark https://github.com/con/duct/pull/110/ as a "minor" update. (Its actually a breaking change, but I don't want auto to go to 1.0.0). When that merges, hopefully we see the magic!

jwodder:
@asmacdo 

* I see you're using the default `auto` label names of "major", "minor", and "patch".  If you ever enable Dependabot in this repository, this will lead to problems, so you may want to rename the labels to something like "semver-major", "semver-minor", and "semver-patch".  See https://github.com/datalad/datalad-installer/issues/175 for more information.
* `release.yml` is currently configured to only run on workflow dispatch, i.e., manually, rather than when a PR is merged.  Is that intentional?
* `.autorc` currently sets `onlyPublishWithReleaseLabel` to `true`, so releases require a PR to be marked with "release", and the "skip-release" label is unnecessary.  Is that intentional?

asmacdo:
ty @jwodder

> release.yml is currently configured to only run on workflow dispatch, i.e., manually, rather than when a PR is merged. Is that intentional?

No that was not intentional, good catch.

> .autorc currently sets onlyPublishWithReleaseLabel to true, so releases require a PR to be marked with "release", and the "skip-release" label is unnecessary. Is that intentional?

I prefer explicitly ask for a release, I've deleted the skip-release label

I've updated the auto labels to use the semver prefix and https://github.com/con/duct/labels to match

yarikoptic (approved):
if you think it is ready -- fire away and see if works! ;)

Issue #114, opened 2024-07-31
Conda build pipeline fails for OSX


asmacdo:
Conda build pipeline is failing on OSX
```
>       memory_total = os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES")
E       ValueError: unrecognized configuration name
```

If I understand correctly, on mac this should be _SC_PAGESIZE and _SC_PHYS_PAGES

IIUC on OSX: 

```
 # E       ValueError: unrecognized configuration name
os.sysconf("SC_PAGE_SIZE")   
os.sysconf("SC_PHYS_PAGES") 
 # should be good?
os.sysconf("_SC_PAGESIZE")  
os.sysconf("_SC_PHYS_PAGES") 
```

@jwodder you have a mac right? Could you sanity check ^ for me?

In the test pipeline stderr I am also seeing https://github.com/con/duct/issues/82
```
ps: illegal option -- s
```

@yarikoptic IMO we should drop support for OSX for now, proceed with the conda-build for linux only. We can address OSX and then cut a new release. wdyt?

jwodder:
@asmacdo The correct expressions on macOS (It hasn't been OS X since 2016) are `os.sysconf("SC_PAGESIZE")` and `os.sysconf("SC_PHYS_PAGES")`.

asmacdo:
@jwodder hmm. thats odd right? The build info looks like it is using up-to-date macOS

2024-07-31T18:48:57.1079870Z ##[section]Starting: Initialize job
2024-07-31T18:48:57.1082990Z Agent name: 'Azure Pipelines 83'
2024-07-31T18:48:57.1083420Z Agent machine name: 'Mac-1722450377722'
2024-07-31T18:48:57.1083640Z Current agent version: '3.242.1'
2024-07-31T18:48:57.1117850Z ##[group]Operating System
2024-07-31T18:48:57.1118100Z macOS
2024-07-31T18:48:57.1118210Z 12.7.5
2024-07-31T18:48:57.1118320Z 21H1222
2024-07-31T18:48:57.1118440Z ##[endgroup]
2024-07-31T18:48:57.1118570Z ##[group]Runner Image
2024-07-31T18:48:57.1118730Z Image: macos-12
2024-07-31T18:48:57.1118880Z Version: 20240721.1
2024-07-31T18:48:57.1119200Z Included Software: https://github.com/actions/runner-images/blob/macos-12/20240721.1/images/macos/macos-12-Readme.md
[...]

PR #115, opened 2024-07-31
SC_PAGESIZE should work on macOS and Linux

asmacdo:
Fixes: https://github.com/con/duct/issues/114


asmacdo:
I'm not clear why the tests passed before this change, but failed on in the conda build pipeline, both are using ` 12.7.5`

jwodder (approved):

PR #123, opened 2024-08-09
Introduce dedicated versioning for the schema

yarikoptic:
So later on we could add checks or even some logic difference based on the schema of the records etc

TODOs
- [ ] make sure saved in all produced .json files (I guess for samples files would need dedicated record with versions information)
- [ ] extend test to ensure that it is saved in all files needing it

### DIFF
--- a/src/con_duct/__init__.py
+++ b/src/con_duct/__init__.py
@@ -1,1 +1,2 @@
 __version__ = "0.1.1"
+__schema_version__ = "0.1.0"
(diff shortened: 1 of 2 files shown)


asmacdo:
> make sure saved in all produced .json files (I guess for samples files would need dedicated record with versions information)
 
This is what i wasn't sure about. currantly the samples (usage.json) file is a set of NLJSON records. It feels weird to have the file start with a record of a completely different type. Maybe we just record the schema version in the info.json, and then when reading the usage.json we also consult info.json?

Issue #116, opened 2024-08-07
store size (bytes) and time (sec) in int or float and in base unit (no killa hours etc)


yarikoptic:
ie. now we have

```
    "wall_clock_time": "818.542 sec",
    "peak_rss": "135928 KiB",
    "average_rss": "135969.137 KiB",
    "peak_vsz": "160752 KiB",
    "average_vsz": "165552.237 KiB",
```

but should be like

```
    "wall_clock_time": 818.542,
    "peak_rss": 139190272,
...
```

even for any kind of "percents" such as

```
    "peak_pmem": "0.2%",
    "average_pmem": "0.199%",
```

should be just

```
    "peak_pmem": 0.2,
    "average_pmem": 0.199,
```

asmacdo:
This should block any work on duct-suite, since this will be a backwards incompatible change. 

@yarikoptic Do you think it would be worth versioning the schema of the output files separately from the duct package?

yarikoptic:
versioning of schema -- sounds a little like overkill at this stage but I would not fight it.
As it is one package, I would not say it would block, it would require more changes (in affected suite components) when done so indeed should be done first and since very simple to do AFAIK, should be done fast ;)

PR #125, opened 2024-08-12
Remove units for machine readability

asmacdo:
Fixes https://github.com/con/duct/issues/116
Fixes https://github.com/con/duct/issues/128

Note: I put KiB into the field names of the execution summary, because `ps` uses KiB, this seemed cleaner than modifying the values into bytes.

```
$ duct -p ZZZZ_ --r-i 0.1 --s-i 0.01 --clobber sleep 1

duct is executing sleep 1...
Log files will be written to ZZZZ_
Exit Code: 0
Command: sleep 1
Log files location: ZZZZ_
Wall Clock Time: 1.001 sec
Memory Peak Usage (RSS) KiB: 1600
Memory Average Usage (RSS) KiB: 1561.905
Virtual Memory Peak Usage (VSZ) KiB: 221584
Virtual Memory Average Usage (VSZ) KiB: 216308.190
Memory Peak Percentage: 0.0%
Memory Average Percentage: 0.000
CPU Peak Usage: 0.0
Average CPU Usage: 0.000
Samples Collected: 42
Reports Written: 11

$ cat ZZZZ_info.json|jq

{
  "command": "sleep 1",
  "system": {
    "uid": "austin",
    "memory_total": 33336778752,
    "cpu_total": 20
  },
  "env": {},
  "gpu": null,
  "duct_version": "0.1.1",
  "execution_summary": {
    "exit_code": 0,
    "command": "sleep 1",
    "logs_prefix": "ZZZZ_",
    "wall_clock_time": "1.001",
    "peak_rss_kib": "1760",
    "average_rss_kib": "1760.000",
    "peak_vsz_kib": "221584",
    "average_vsz_kib": "221584.000",
    "peak_pmem": "0.0%",
    "average_pmem": "0.000",
    "peak_pcpu": "0.0",
    "average_pcpu": "0.000",
    "num_samples": 41,
    "num_reports": 10
  }
}
```


asmacdo:
@yarikoptic it kinda bothers me that "memory_total" is bytes, but the samples are KiB, but its fine. WDYT

src/con_duct/__main__.py
```diff
@@ -345,24 +345,24 @@ def execution_summary(self) -> dict[str, Any]:
             "exit_code": self.process.returncode,
             "command": self.command,
             "logs_prefix": self.log_paths.prefix,
-            "wall_clock_time": f"{self.elapsed_time:.3f} sec",
-            "peak_rss": (
-                f"{self.max_values.total_rss} KiB"
+            "wall_clock_time": f"{self.elapsed_time:.3f}",
+            "peak_rss_kib": (
```
yarikoptic, line 349: I think we better just report in bytes. Check ps options or convert if needed from kbs

src/con_duct/__main__.py
```diff
@@ -345,46 +345,46 @@ def execution_summary(self) -> dict[str, Any]:
...
+            "wall_clock_time": f"{self.elapsed_time:.3f}",
             "peak_rss": (
-                f"{self.max_values.total_rss} KiB"
+                f"{self.max_values.total_rss}"
                 if self.max_values.stats
                 else "unknown"
             ),
             "average_rss": (
-                f"{self.averages.rss:.3f} KiB"
+                f"{int(self.averages.rss)}"
```
asmacdo, line 355: if we are reporting bytes, i think `int` is preferable to `float`

yarikoptic: Yes, but might be worth checking if a number at all first

yarikoptic:
Didn't forget to add a test

asmacdo:
@yarikoptic I'm not sure what kind of test you had in mind, but since there are other places earlier in the code that would go wrong if those values weren't numbers.

I figured it would be best to just add some more assertions into the runtime, and add validation where we first get the values from `ps`. 

lmk if this isn't what you had in mind.

asmacdo:
@yarikoptic done pushing ready for re-review

src/con_duct/__main__.py
```diff
@@ -96,6 +96,21 @@ def max(self, other: ProcessStats) -> ProcessStats:
...
+        if not isinstance(self.pcpu, (float, int)):
+            raise TypeError(f"Expected 'pcpu' to be of type 'float' or 'int', got {type(self.pcpu).__name__}")
+        if not isinstance(self.pmem, (float, int)):
+            raise TypeError(f"Expected 'pmem' to be of type 'float' or 'int', got {type(self.pmem).__name__}")
+        if not isinstance(self.rss, int):
+            raise TypeError(f"Expected 'rss' to be of type 'int', got {type(self.rss).__name__}")
+        if not isinstance(self.vsz, int):
+            raise TypeError(f"Expected 'vsz' to be of type 'int', got {type(self.vsz).__name__}")
+        if not isinstance(self.timestamp, str):
+            raise TypeError(f"Expected 'timestamp' to be of type 'str', got {type(self.timestamp).__name__}")
```
yarikoptic, line 112: I really dislike duplication. Would be better to have something like
```suggestion
        for f, types in (('pcpu', (float, int)), ...):
            value = getattr(self, f)
            if not isinstance(value, types):
                raise TypeError(f"Expected '{f}' to be of type {' or '.join(map(repr(types)))}, got {type(value).__name__}")
```

but also wondered if there is some runtime way to take those from type annotations to avoid duplication here or just do "runtime type validation" (like with pydantic)?

src/con_duct/__main__.py
```diff
@@ -345,46 +367,46 @@ def execution_summary(self) -> dict[str, Any]:
             "exit_code": self.process.returncode,
             "command": self.command,
             "logs_prefix": self.log_paths.prefix,
-            "wall_clock_time": f"{self.elapsed_time:.3f} sec",
+            "wall_clock_time": f"{self.elapsed_time:.3f}",
```
yarikoptic, line 370: why is it still formatted to str here and below -- shouldn't we just leave original numerics here?

test/test_report.py
```diff
@@ -89,3 +91,90 @@ def test_averages_three_samples() -> None:
...
+
+def test_process_stats_incorrect_ts_type() -> None:
+    with pytest.raises(TypeError):
+        ProcessStats(
+            pcpu=1.1,
+            pmem=1.0,
+            rss=1025,
+            vsz=1024,
+            timestamp=1  # type: ignore[arg-type]
+        )
```
yarikoptic, line 170: duplication makes it hard(er) to see the critical difference (even though it is in the name and even in type annotation comment).  IMHO would have been much better done via `@pytest.mark.parametrize` having some common `kwargs` dict which you `.update` with the specific parameter (specific `attribute=value`) and have a generic construct. Makes then really easy to see what that particular test is doing, that it is just 1 test and different parametrizations.

yarikoptic (changes requested):
left individual comments, please reduce code duplication and possibly return original numerics in the record?

src/con_duct/__main__.py
asmacdo: Seems reasonable to me, this is there just for rounding, otherwise its unhelpfully "precise"

` "wall_clock_time": 1.0147314071655273`

I'll update to use  `round`, which would drop the unnecessary quotes. 

` "wall_clock_time": 1.015`

src/con_duct/__main__.py
asmacdo: Considering this approach makes me wonder if we are going in the wrong direction.

Instead of blowing up if we get an invalid value, it would probably be more helpful to just record NoneType. (I'm thinking of pandas NaN as an example)

In the case where we get invalid information, we should just mark it as a broken sample, log the error, but continue with the execution

yarikoptic: we could. your call

src/con_duct/__main__.py
yarikoptic: rounding etc should be done by "rendering"/styling function. Is this a rendering function or that is a general record/model composition?

asmacdo: addressed in latest push

src/con_duct/__main__.py
asmacdo: It would be pretty obnoxious to deal with the possibility of other types all over the code to cover a hypothetical case, so for now I went with the simplest runtime type validation I could do. We can come back to it if we get weird data from `ps`

yarikoptic (changes requested):
minor recommendation on clearer variable naming and one reference for that "default" (now is also a separate issue so could be done separately).

src/con_duct/__main__.py
```diff
@@ -158,6 +169,9 @@ class Averages:
     num_samples: int = 0
 
     def update(self: Averages, other: Sample) -> None:
+        assert_num(
+            other.total_rss, other.total_vsz, other.total_pmem, other.total_pcpu
+        )
```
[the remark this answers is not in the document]
asmacdo: Going to address this in the next one, dropping the release tag here, will add over there.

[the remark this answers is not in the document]
asmacdo: https://github.com/con/duct/pull/135

Issue #133, opened 2024-08-14
Replace ad-hoc "unknown" string in records with numbers to None


yarikoptic:
ref eg. https://github.com/con/duct/pull/125/files#diff-e3756666bbaae702279bc0504178cc15b9a612126ac8646e6156ad6f74d9d33aR373 

just make all the values which could be None, to default None in the data classes. Then there would be no need for duplicate `if else` all over.

PR #135, opened 2024-08-14
Use None rather than 0 prior to measurement

asmacdo:
Fixes https://github.com/con/duct/issues/133

If we have not measured, 0 is incorrect.

We also substitute `unknown` for None when printing for the user

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -13,7 +13,7 @@ import subprocess
 import sys
 import threading
 import time
-from typing import Any, IO, Optional, TextIO
+from typing import IO, Any, Optional, TextIO
 from . import __version__
 
 ENV_PREFIXES = ("PBS_", "SLURM_", "OSG")
@@ -406,7 +406,10 @@ class Report:
 
     @cached_property
     def execution_summary_formatted(self) -> str:
-        return {k: "unknown" if v is None else v for k, v in self.execution_summary.items()}
+        human_readable = {
+            k: "unknown" if v is None else v for k, v in self.execution_summary.items()
+        }
+        return self.summary_format.format_map(human_readable)
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


src/con_duct/__main__.py
```diff
@@ -162,21 +162,36 @@ def prepare_paths(self, clobber: bool, capture_outputs: Outputs) -> None:
...
+            self.pcpu = other.total_pcpu
+        else:
+            assert self.rss is not None
+            assert self.vsz is not None
+            assert self.pmem is not None
+            assert self.pcpu is not None
+            assert other.total_rss is not None
+            assert other.total_vsz is not None
+            assert other.total_pmem is not None
+            assert other.total_pcpu is not None
```
yarikoptic, line 189: just a note since not relevant to this PR: I have odd feeling here about code that this `self` is having assumptions about `other` (Sample)

src/con_duct/__main__.py
```diff
@@ -162,21 +162,36 @@ def prepare_paths(self, clobber: bool, capture_outputs: Outputs) -> None:
...
+            self.num_samples += 1
+            self.rss += (other.total_rss - self.rss) / self.num_samples
+            self.vsz += (other.total_vsz - self.vsz) / self.num_samples
+            self.pmem += (other.total_pmem - self.pmem) / self.num_samples
+            self.pcpu += (other.total_pcpu - self.pcpu) / self.num_samples
```
yarikoptic, line 194: just another note: I feel odd about this math here... unrelated to this PR though, so just a note.

src/con_duct/__main__.py
```diff
@@ -406,7 +409,8 @@ def dump_json(self) -> str:
         )
 
     def print_summary(self) -> None:
-        print(self.summary_format.format_map(self.execution_summary))
+        human_readable = {k: "unknown" if v is None else v for k, v in self.execution_summary.items()}
+        print(self.summary_format.format_map(human_readable))
```
yarikoptic, line 413: is there a unittest which would test this function that logic of conversion to `human_readable` is correct?

yarikoptic (commented):
two notes and one question (which might need a unittest being added)

src/con_duct/__main__.py
asmacdo: there is now ;)

PR #136, opened 2024-08-14
RF+ENH: output messages to stderr not stdout + move printing out of "controllers/models"

yarikoptic:
Collection of two commits

- Remove unused/untested print_max_values
-   Most of the tools which operate over some other tool/streams (e.g. pv, datalad
    run, etc) strive to not interfer with stdout so then they could be used
    within pipe constructs.  If such wrapper tool outputs to stdout, then it could
    not be used within a pipe, e.g. in
    
          duct my_amportant_compute | grep "a"
    
    construct.  The same goes for git and git-annex, datalad, etc, unless it is
    explicitly a command intended to display something (e.g. "datalad wtf").
    
    That is why important to not just "print" to stdout but use e.g. stderr which
    is not piped through.  Relates also to issue on logging which should default to
    stderr not stdout.

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -352,10 +352,6 @@ class Report:
                 json.dumps(self.current_sample.for_json()) + "\n"
             )
 
-    def print_max_values(self) -> None:
-        for pid, maxes in self.max_values.stats.items():
-            print(f"PID {pid} Maximum Values: {asdict(maxes)}")
-
     @cached_property
     def execution_summary(self) -> dict[str, Any]:
         return {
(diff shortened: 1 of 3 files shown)


src/con_duct/__main__.py
```diff
@@ -330,7 +328,9 @@ def collect_sample(self) -> Sample:
             )
             for line in output.splitlines()[1:]:
                 if line:
-                    pid, pcpu, pmem, rss_kib, vsz_kib, etime, cmd = line.split(maxsplit=6)
+                    pid, pcpu, pmem, rss_kib, vsz_kib, etime, cmd = line.split(
+                        maxsplit=6
+                    )
```
yarikoptic, line 333: these changes are unrelated but the result of not having pre-commit "install"ed/used to `black` the code.

PR #139, opened 2024-08-14
RF: no shebang since file is no longer can be executed

yarikoptic:
there is relative import so package is required


yarikoptic:
kinda minor and agreed upon, so let's merge to not linger

PR #140, opened 2024-08-15
RF: Add logging, dissolve duct_print (INFO level), add CLI option -l, dissolve --quiet

### LINKED ISSUE #134: Add logger

yarikoptic:
When we have logging, and it also goes to stderr, there is really no need for some dedicated "duct_print". Now everything can be controlled with -l and no additional --quiet is needed (good that it was not demoed in https://blog.datalad.org/ post).

I also added a few lgr.debug level statements.

I delayed setup of logging.basicConfig until `main` so it might be more coherent later on with desire to make this whole thing into a reusable Python module

Closes #134

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -17,8 +17,6 @@ from typing import IO, Any, Optional, TextIO
 from . import __version__
 
 lgr = logging.getLogger("con-duct")
-# Default log level is DEBUG, but can be overridden by setting DUCT_LOG_LEVEL
-# and later set via CLI argument
 DEFAULT_LOG_LEVEL = os.environ.get("DUCT_LOG_LEVEL", "INFO").upper()
 
 ENV_PREFIXES = ("PBS_", "SLURM_", "OSG")
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -680,6 +680,7 @@ def execute(args: Arguments) -> int:
 
     Returns exit code of the executed process.
     """
+    lgr.setLevel(args.log_level)
     log_paths = LogPaths.create(args.output_prefix, pid=os.getpid())
     log_paths.prepare_paths(args.clobber, args.capture_outputs)
     stdout, stderr = prepare_outputs(args.capture_outputs, args.outputs, log_paths)
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -19,7 +19,7 @@ from . import __version__
 lgr = logging.getLogger("con-duct")
 # Default log level is DEBUG, but can be overridden by setting DUCT_LOG_LEVEL
 # and later set via CLI argument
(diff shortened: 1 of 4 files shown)


src/con_duct/__main__.py
```diff
@@ -359,7 +365,8 @@ def collect_sample(self) -> Optional[Sample]:
                             timestamp=datetime.now().astimezone().isoformat(),
                         ),
                     )
-        except subprocess.CalledProcessError:  # when session_id has no processes
+        except subprocess.CalledProcessError as exc:  # when session_id has no processes
+            lgr.debug("Error collecting sample: %s", str(exc))
             return None
```
asmacdo, line 370:
```suggestion
        except subprocess.CalledProcessError as exc: 
            lgr.debug("Could not collect sample (process may have finished): %s", str(exc))
            return None
```
I know its just debug, but since this will happen almost every run we should make it sound not scary to unsuspecting users ;)

src/con_duct/__main__.py
```diff
@@ -15,6 +16,11 @@
 from typing import IO, Any, Optional, TextIO
 from . import __version__
 
+lgr = logging.getLogger("con-duct")
+# Default log level is DEBUG, but can be overridden by setting DUCT_LOG_LEVEL
+# and later set via CLI argument
+DEFAULT_LOG_LEVEL = os.environ.get("DUCT_LOG_LEVEL", "DEBUG").upper()
```
asmacdo, line 22:
```suggestion
DEFAULT_LOG_LEVEL = os.environ.get("DUCT_LOG_LEVEL", "INFO").upper()
```
DEBUG is too low for default IMO

src/con_duct/__main__.py
```diff
@@ -15,6 +16,11 @@
 from typing import IO, Any, Optional, TextIO
 from . import __version__
 
+lgr = logging.getLogger("con-duct")
+# Default log level is DEBUG, but can be overridden by setting DUCT_LOG_LEVEL
+# and later set via CLI argument
```
asmacdo, line 21:
```suggestion
```
pretty obvious imo no need for comment

asmacdo (approved):
some suggestions, feel free to accept as you see fit and merge when ready.

FWIW https://blog.datalad.org/posts/intro-duct-tion/ will be out of date (uses `--quiet`) lets fix that before we release

src/con_duct/__main__.py
yarikoptic: Ok
No need for str though

src/con_duct/__main__.py
yarikoptic: Right!

yarikoptic:
Dang, is searched for it, I guess for silent not quiet. I will add deprecation warning when get to laptop

src/con_duct/__main__.py
asmacdo: Since its set to DEBUG, i dont see a problem leaving the exc in, it could be useful if ps fails for some other reason

PR #141, opened 2024-08-15
RF+BF: make explicit wall_clock_time separate from elapsed_time

### LINKED ISSUE #127: wall clocktime should be more "precise"
```
❯ time sleep 2
sleep 2  0.00s user 0.00s system 0% cpu 2.003 total

❯ duct sleep 2
duct is executing sleep 2...
Log files will be written to .duct/logs/2024.08.13T10.31.01-2450984_
Exit Code: 0
Command: sleep 2
Log files location: .duct/logs/2024.08.13T10.31.01-2450984_
Wall Clock Time: 3.814 sec
```

so it seems it adds almost 2 seconds to actual 2 seconds of run of `sleep 2`.  I guess it should monitor the end of the process more "closely" I guess.

yarikoptic:
Also moved stop_time as close to .wait as possible.  Now with such changes we get some reasonable precision in reported wall clock time:

    ❯ duct sleep 1 2>&1 | grep Wall
    Wall Clock Time: 1.002 sec
    ❯ duct sleep 0.1 2>&1 | grep Wall
    Wall Clock Time: 0.102 sec

- Closes #127

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -280,7 +280,7 @@ class Report:
         self.summary_format: str = summary_format
         self.clobber = clobber
         # Defaults to be set later
-        self.start_time = None
+        self.start_time: float | None = None
         self.process = process
         self.session_id: int | None = None
         self.gpus: list[dict[str, str]] | None = None
@@ -299,7 +299,7 @@ class Report:
 
     @property
     def elapsed_time(self) -> float:
-        assert self.start_time
+        assert self.start_time is not None
         return time.time() - self.start_time
 
     @property
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -389,8 +389,7 @@ class Report:
 
     @property
     def execution_summary(self) -> dict[str, Any]:
-        # prepare the base, but enrich if we did get process
-        # running
+        # prepare the base, but enrich if we did get process running
         return {
             "exit_code": self.process.returncode if self.process else None,
             "command": self.command,
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -718,7 +718,7 @@ def execute(args: Arguments) -> int:
 
     full_command = " ".join([str(args.command)] + args.command_args)
     files_to_close = [stdout_file, stdout, stderr_file, stderr]
-    
+
     report = Report(
         args.command,
         args.command_args,
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -7,6 +7,7 @@ from enum import Enum
 from functools import cached_property
 import json
 import logging
+import math
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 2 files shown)


src/con_duct/__main__.py
```diff
@@ -265,38 +266,50 @@ def __init__(
...
+        process: subprocess.Popen | None = None,
     ) -> None:
-        self.start_time = time.time()
         self._command = command
         self.arguments = arguments
-        self.session_id = session_id
+        self.log_paths = log_paths
+        self.summary_format: str = summary_format
+        self.clobber = clobber
+        # Defaults to be set later
```
asmacdo, line 279:
```suggestion
        # Defaults to be set later by argparse
```

test/test_report.py
```diff
@@ -149,10 +149,14 @@ def test_execution_summary_formatted(
     mock_popen: mock.MagicMock, mock_log_paths: mock.MagicMock
 ) -> None:
     mock_log_paths.prefix = "mock_prefix"
-    report = Report(
-        "_cmd", [], None, mock_popen, mock_log_paths, EXECUTION_SUMMARY_FORMAT, False
-    )
+    report = Report("_cmd", [], None, EXECUTION_SUMMARY_FORMAT, clobber=False)
+    # It should not crash and it would render even where no wallclock time yet
+    assert "wall clock time: nan" in report.execution_summary_formatted.lower()
```
asmacdo, line 154: Do you think its safe to use this cached property this early? Previously, the cached properties weren't used until the end and the data was stable.

src/con_duct/__main__.py
```diff
@@ -265,38 +266,50 @@ def __init__(
...
+        assert self.start_time
         return time.time() - self.start_time
 
+    @property
+    def wall_clock_time(self) -> Optional[float]:
+        if self.start_time is None:
+            return math.nan
+        if self.end_time is None:
+            # if no end_time -- must be still ongoing
+            # Cannot happen ATM but could in "library mode" later
```
asmacdo, line 308:
```suggestion
            # Cannot happen ATM but could in "library mode" later
            # TODO: log this
```

test/test_report.py
yarikoptic: good catch. Let's make them into regular property!

yarikoptic: done

src/con_duct/__main__.py
```diff
@@ -370,13 +382,15 @@ def write_subreport(self) -> None:
                 json.dumps(self.current_sample.for_json()) + "\n"
             )
 
-    @cached_property
+    @property
     def execution_summary(self) -> dict[str, Any]:
+        # prepare the base, but enrich if we did get process
+        # running
```
asmacdo, line 388:
```suggestion
        # prepare the base, but enrich if we did get process running
```

src/con_duct/__main__.py
asmacdo: I misunderstood why this would happen in library mode (this function would be called prior to end), so this isnt a critical/weird thing to happen.

Issue #150, opened 2024-08-20
Improve "rendering" of summary and possibly fix collection of "percenages" and memory usage


yarikoptic:
ATM as of 0.1.1 (reported, really v0.1.1-58-g0508635) we have

```
Wall Clock Time: 101.877 sec
Memory Peak Usage (RSS): 14909440 bytes
Memory Average Usage (RSS): 14888140.799999997 bytes
Virtual Memory Peak Usage (VSZ): 250023936 bytes
Virtual Memory Average Usage (VSZ): 247739432.95999998 bytes
Memory Peak Percentage: 250023936
Memory Average Percentage: 247739432.95999998
CPU Peak Usage: 200.0
Average CPU Usage: 3.5020000000000002
Samples Collected: 100
Reports Written: 2
```

- I really don't know how we got into those floating point memory sizes (bytes) -- should be ints all around.
- For CPU Usage add ` %` at the end and trim to `.3f` .
- For "Percentage" also add ` %` and trim to `.3f`.

I also spotted that memory average percentages  make no sense whatsoever! Fix compute I guess.

- I guess also add a test to ensure that values make sense, may be even in code check that "Memory ... Percentages" do not go over e.g. 100  
edit: that was from `duct duct sleep 100`. Full report is in #60

Issue #82, opened 2024-06-10
Portability issue: `ps` on macOS does not have `-s` option


jwodder:
[`duct` invokes `ps` with the `-s` option](https://github.com/con/duct/blob/7e26b20872bc0752cfc9016c069a6ddb3aaf6242/src/duct.py#L130-L134) for filtering by session ID, but the version of `ps` available on macOS by default (as of macOS Sonomoa 14.5 on an Intel chip) does not have this option (or, seemingly, any way to directly filter by session).

asmacdo:
Thats unfortunate-- `ps -s` is the backbone of how this whole thing works :/

jwodder:
One possible option for working around this would be to give up on the idea of "no dependencies" (not sure why you're prioritizing that to begin with) and using https://github.com/giampaolo/psutil instead of `ps`.

asmacdo:
Without dependencies, the script is runnable even without "installing" by just executing the script directly. @yarikoptic suggested this would be helpful in tightly controlled HPC environments.

jwodder:
Surely such environments would still allow installation inside a virtualenv, though?

yarikoptic:
https://github.com/giampaolo/psutil looks interesting ... but that one also does not seems to have a concept for "session" (upon a [quick search](https://github.com/search?q=repo%3Agiampaolo%2Fpsutil%20session&type=code)). 

I wonder what would be the preferable way then to track the children processes using `psutil`?

NB smells like we might need some pluggable/switchable setup to allow for different ways to track children processes if we decide to go cross-platform which in general would be great to have.

yarikoptic:
@jwodder could you please research on the concept of "sessions" under OSX and if `ps` could be used then to get processes for a specific session?

jwodder:
@yarikoptic 

> could you please research on the concept of "sessions" under OSX

[They exist.](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getsid.2.html)

> and if ps could be used then to get processes for a specific session?

Based on the `ps` manpage, the only viable option seems to be to use the `-o` or `-O` option to include `sess` (session ID) in the output while listing all processes and then filter the results inside `duct`.

I really think you should reconsider your aversion to using `psutil` as a dependency.

jwodder:
Is using sessions rather than just process groups even the best idea here?  Sessions seem to be intended for things like login sessions.

yarikoptic:
[...]

PR #151, opened 2024-08-20
Fix typo in pmem calculation

asmacdo:
Fixes https://github.com/con/duct/issues/150

Also
 - use floats rather than ints for 0
 - add % to human readable summary

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-12
+          # - macos-12
           - macos-latest
           # - windows-latest
           - ubuntu-latest
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         os:
           # - macos-12
-          - macos-latest
+          # - macos-latest
           # - windows-latest
           - ubuntu-latest
         python-version:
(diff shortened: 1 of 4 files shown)


yarikoptic (changes requested):
add a test . 
could mimic original invocation which observed that odd behavior: run `duct duct sleep 1` with some short periods for sampling/report

yarikoptic:
tests aren't happy , hadn't looked in detail yet.

asmacdo:
All the MacOS tests are failing because we cannot collect samples. The SANITY commit is not intended to be merged, its to show that there was plenty of time to collect a sample. This is related to https://github.com/con/duct/issues/82

asmacdo:
From one of the failing mac runs *Samples Collected: 0*, most likely because `ps -s` isnt an option.
```
Wall Clock Time: 1.139 sec
Memory Peak Usage (RSS): unknown bytes
Memory Average Usage (RSS): unknown bytes
Virtual Memory Peak Usage (VSZ): unknown bytes
Virtual Memory Average Usage (VSZ): unknown bytes
Memory Peak Percentage: unknown%
Memory Average Percentage: unknown%
CPU Peak Usage: unknown%
Average CPU Usage: unknown%
Samples Collected: 0
Reports Written: 0
```

asmacdo:
image

The tests didnt run, looks like a partial outage

.github/workflows/test.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-12
+          # - macos-12
           - macos-latest
```
yarikoptic, line 30:
```suggestion
          # - macos-latest
```

since unlikely to work there either

Issue #126, opened 2024-08-13
Add hostname to _info.json:system


yarikoptic:
we already have `uid`. Together with `hostname` would be useful to figure out where actually compute has happened.

PR #153, opened 2024-08-20
Collect hostname in sys info

asmacdo:
Fixes https://github.com/con/duct/issues/126

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,6 +10,7 @@ deps =
     pytest-cov
 commands =
     pytest {posargs} test
+passenv = USER
 
 [testenv:lint]
 deps =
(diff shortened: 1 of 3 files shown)


yarikoptic:
There should be some rudimentary testing added somewhere that we do have such a field and it has a value (not None or empty string)

yarikoptic (approved):

tox.ini
```diff
@@ -10,6 +10,7 @@ deps =
     pytest-cov
 commands =
     pytest {posargs} test
+passenv = USER
```
yarikoptic, line 13: what happens if not passed?

asmacdo: User is None when run in tic

Issue #121, opened 2024-08-08
Add an introductory blurb to README


asmacdo:
Lets do this by expanding the introduction of --help

PR #158, opened 2024-08-21
Improve helptext top-level description

asmacdo:
Fixes https://github.com/con/duct/issues/121

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,7 @@ execute a command, particularly in an HPC environment.
 
 Resource usage is determined by polling, (at a sample-interval).
 During execution, duct produces a JSON lines (see https://jsonlines.org) file
-with one data
-point recorded for each report, (of frequency report-interval).
+with one data point recorded for each report, (of frequency report-interval).
 
 environment variables:
   Many duct options can be configured by environment variables (which are
--- a/README.md
+++ b/README.md
@@ -36,12 +36,11 @@ duct is a lightweight wrapper that collects execution data for an arbitrary
 command.  Execution data includes execution time, system information, and
 resource usage statistics of the command and all its child processes. It is
 intended to simplify the problem of recording the resources necessary to
-execute a command to improve reproducibility, particularly in an HPC
-environment.
+execute a command, particularly in an HPC environment.
 
-Resource usage is determined by polling, (of frequency sample_interval).
-During execution, duct produces a Newline Deliminated JSON file with one data
-point recorded for each report, (of frequency report-interval).
+Resource usage is determined by polling (at a sample-interval).
+During execution, duct produces a JSON lines (see https://jsonlines.org) file with one data
+point recorded for each report (at a report-interval).
 
 environment variables:
   Many duct options can be configured by environment variables (which are
--- a/README.md
+++ b/README.md
@@ -33,8 +33,27 @@ usage: duct [-h] [--version] [-p OUTPUT_PREFIX]
             command [command_args ...] ...
 
 duct is a lightweight wrapper that collects execution data for an arbitrary
-command. Execution data includes execution time, system information, and
-resource usage statistics of the command and all its child processes.
+command.  Execution data includes execution time, system information, and
+resource usage statistics of the command and all its child processes. It is
+intended to simplify the problem of recording the resources necessary to
+execute a command to improve reproducibility, particularly in an HPC
+environment.
+
+Resource usage is determined by polling, (of frequency sample_interval).
+During
+execution, duct produces a Newline Deliminated JSON file with one data point
+recorded for each report, (of frequency report-interval).
+
+environment variables:
+  Many duct options can be configured by environment variables (which are
+  overridden by command line options).
+
+  DUCT_LOG_LEVEL: see --log-level
+  DUCT_OUTPUT_PREFIX: see --output-prefix
+  DUCT_SUMMARY_FORMAT: see --summary-format
+  DUCT_SAMPLE_INTERVAL: see --sample-interval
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -32,7 +32,9 @@ usage: duct [-h] [--version] [-p OUTPUT_PREFIX]
             [-t {all,system-summary,processes-samples}]
             command [command_args ...] ...
 
-Gathers metrics on a command and all its child processes.
+duct is a lightweight wrapper that collects execution data for an arbitrary
+command. Execution data includes execution time, system information, and
+resource usage statistics of the command and all its child processes.
```
yarikoptic, line 37: should mention that it samples that process and children periodically, so that information could be used to later analyze/plot command's resources consumption.  And that it might be useful to declare needed resources for allocation on HPC tasks, or for troubleshooting.  Also note that it stores all the standard output and error into external files.
Do we have somewhere listing of env vars to configure (may be in README for now)? should refer to that description here too IMHO.

asmacdo: > should mention that it samples that process and children periodically, so that information could be used to later analyze/plot command's resources consumption. And that it might be useful to declare needed resources for allocation on HPC tasks, or for troubleshooting. Also note that it stores all the standard output and error into external files.

+1

> Do we have somewhere listing of env vars to configure (may be in README for now)? should refer to that description here too IMHO.

Its not listed anywhere atm, +1 to add here.

asmacdo:
@yarikoptic ready for re-review. 

This does what we wanted, but we lose automatic line wrapping based on terminal width. (datalad --help is similarly limited). IMO its annoying but worth the improved helptext.

README.md
```diff
@@ -32,7 +32,27 @@ usage: duct [-h] [--version] [-p OUTPUT_PREFIX]
             [-t {all,system-summary,processes-samples}]
             command [command_args ...] ...
 
-Gathers metrics on a command and all its child processes.
+duct is a lightweight wrapper that collects execution data for an arbitrary
+command.  Execution data includes execution time, system information, and
+resource usage statistics of the command and all its child processes. It is
+intended to simplify the problem of recording the resources necessary to
+execute a command to improve reproducibility, particularly in an HPC
+environment.
```
yarikoptic, line 40:
```suggestion
execute a command, particularly in an HPC environment.
```

since AFAIK there is little to nothing about reproducibility aspect here.

README.md
```diff
@@ -32,7 +32,27 @@ usage: duct [-h] [--version] [-p OUTPUT_PREFIX]
...
+
+Resource usage is determined by polling, (of frequency sample_interval).
```
yarikoptic, line 42: odd syntax + frequency is a reciprocal to interval

```suggestion
Resource usage is determined by polling (at a sample-interval).
```

note change of `_` to `-` for consistency with other similar use below

README.md
```diff
@@ -32,7 +32,27 @@ usage: duct [-h] [--version] [-p OUTPUT_PREFIX]
...
+During execution, duct produces a Newline Deliminated JSON file with one data
```
yarikoptic, line 43:
```suggestion
During execution, duct produces a JSON lines (see https://jsonlines.org) file with one data
```

since it is not JSON line per se

README.md
```diff
@@ -32,7 +32,27 @@ usage: duct [-h] [--version] [-p OUTPUT_PREFIX]
...
+point recorded for each report, (of frequency report-interval).
```
yarikoptic, line 44: see above

```suggestion
point recorded for each report (at a report-interval).
```

src/con_duct/__main__.py
```diff
@@ -43,6 +44,35 @@
...
+ABOUT_DUCT = """
+duct is a lightweight wrapper that collects execution data for an arbitrary
+command.  Execution data includes execution time, system information, and
+resource usage statistics of the command and all its child processes. It is
+intended to simplify the problem of recording the resources necessary to
+execute a command to improve reproducibility, particularly in an HPC environment.
+
+Resource usage is determined by polling, (of frequency sample_interval).
+During execution, duct produces a Newline Deliminated JSON file with one data
+point recorded for each report, (of frequency report-interval).
```
yarikoptic, line 56: dang, forgot that README is produced from here. See recommended changes above

yarikoptic (changes requested):
minor wording recommendations/fixes

src/con_duct/__main__.py
asmacdo: We don't have to include the env vars here, the readme is only updated between the markers, if you prefer to move them out of the helptext

asmacdo: oh i see, you meant apply the changes to the source, not the destination. got it-- they are applied now

src/con_duct/__main__.py
```diff
@@ -43,6 +44,35 @@
...
+ABOUT_DUCT = """
+duct is a lightweight wrapper that collects execution data for an arbitrary
+command.  Execution data includes execution time, system information, and
+resource usage statistics of the command and all its child processes. It is
+intended to simplify the problem of recording the resources necessary to
+execute a command, particularly in an HPC environment.
+
+Resource usage is determined by polling, (at a sample-interval).
+During execution, duct produces a JSON lines (see https://jsonlines.org) file
+with one data point recorded for each report, (of frequency report-interval).
```
yarikoptic, line 56:
```suggestion
Resource usage is determined by polling (at a sample-interval).
During execution, duct produces a JSON lines (see https://jsonlines.org) file
with one data point recorded for each report (at a report-interval).
```

yarikoptic (changes requested):
more of recommended changes to be transferred to .py

PR #159, opened 2024-08-21
Add log level NONE and deprecate quiet

asmacdo:
Fixes https://github.com/con/duct/issues/154

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -513,7 +513,7 @@ class Arguments:
             "--log_level",
             default=DEFAULT_LOG_LEVEL,
             choices=("NONE", "CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"),
-            help="Log level from duct operation, use NONE to disable duct output to stderr.",
+            help="Level of log output to stderr, use NONE to entirely disable.",
         )
         parser.add_argument(
             "-q",
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -512,14 +512,14 @@ class Arguments:
             "-l",
             "--log_level",
             default=DEFAULT_LOG_LEVEL,
-            choices=("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"),
-            help="Log level from duct operation.",
+            choices=("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG", "NONE"),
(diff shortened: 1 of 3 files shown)


src/con_duct/__main__.py
```diff
@@ -512,14 +512,14 @@ def from_argv(
...
-            help="Log level from duct operation.",
+            choices=("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG", "NONE"),
+            help="Log level from duct operation, use NONE to disable duct output to stderr.",
         )
         parser.add_argument(
             "-q",
             "--quiet",
             action="store_true",
-            help="Disable duct logging output (to stderr)",
+            help="[deprecated, use log level NONE] Disable duct logging output (to stderr)",
```
yarikoptic, line 522: it would take longer for people to notice. Ideally there should be some kind of DeprecationWarning if this option is used.  Could you please add one on `if quiet:` ?

asmacdo: Unfortunately no, if --quiet, we have to actually be quiet :)

src/con_duct/__main__.py
```diff
@@ -512,14 +512,14 @@ def from_argv(
             "-l",
             "--log_level",
             default=DEFAULT_LOG_LEVEL,
-            choices=("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"),
-            help="Log level from duct operation.",
+            choices=("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG", "NONE"),
```
yarikoptic, line 515:
```suggestion
            choices=("NONE", "CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"),
```
if to have it ordered from "nothing" to the "most".

asmacdo: its currently order most to None, would you prefer to swap the whole order?

src/con_duct/__main__.py
yarikoptic: hm, makes sense although rubs against my instinct here... ok for now

src/con_duct/__main__.py
```diff
@@ -512,14 +512,14 @@ def from_argv(
...
+            help="Log level from duct operation, use NONE to disable duct output to stderr.",
```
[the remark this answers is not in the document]
asmacdo: i like this better +1

yarikoptic (approved):

PR #167, opened 2024-08-28
Argparse abbreviation affects and breaks cmd args

asmacdo:
Builds on https://github.com/con/duct/pull/166 (merge that first)

This came up while using duct with `mriqc` 

```
duct -p ../duct_out/Z_ --clobber datalad containers-run \
        -n bids-mriqc \
        --input sourcedata \
        --output . \
        '{inputs}' '{outputs}' participant --participant-label 02 -w workdir
```

The `--output` option caused the command to fail:
 `duct: error: ambiguous option: --output could match --output-prefix, --outputs`

### DIFF
--- a/test/test_arg_parsing.py
+++ b/test/test_arg_parsing.py
@@ -49,6 +49,7 @@ def test_abreviation_disabled() -> None:
     """
     try:
         subprocess.check_output(["duct", "ps", "--output"], stderr=subprocess.STDOUT)
+        raise AssertionError("Invocation of 'ps' should have failed")
     except subprocess.CalledProcessError as e:
         assert e.returncode == 1
         assert "duct: error: ambiguous option: --output could match" not in str(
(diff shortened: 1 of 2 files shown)


test/test_arg_parsing.py
```diff
@@ -40,3 +40,18 @@ def test_duct_missing_cmd() -> None:
...
+
+
+def test_abreviation_disabled() -> None:
+    """
+    If abbreviation is enabled, options passed to command (not duct) are still
+    filtered through the argparse and causes problems.
+    """
+    try:
+        subprocess.check_output(["duct", "ps", "--output"], stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
```
yarikoptic, line 52:
```suggestion
        raise AssertionError("Invocation of 'ps' should have failed")
    except subprocess.CalledProcessError as e:
```
or alike?

Issue #148, opened 2024-08-19
reports negative exit code -- should we treat as a feature or "fix up"


yarikoptic:
```
❯ duct strace -t -f -o .tmp/`mdate`-nvidia-smi-strace-1.log nvidia-smi
...
Exit Code: -9
```
whenever
```
❯ strace -t -f -o .tmp/`mdate`-nvidia-smi-strace-2.log nvidia-smi ; echo $?
...
137

```
which is due to https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.exitcode

> If the child terminated due to an exception not caught within [run()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.run), the exit code will be 1. If it was terminated by signal N, the exit code will be the negative value -N.

But isn't it a "Python specific behavior" and generally we should store unsigned ints? WDYT @jwodder ?

asmacdo:
Weird, even with that behavior I was expecting the error to be `-137` not `-9`.

jwodder:
If we're talking about what exit code `duct` should print when the subprocess is killed by a signal, I think "Exit Code: {n}" should be forgone in such circumstances and be replaced by something like "Killed by signal: {signal_name}".

If we're talking about what exit code `duct` should exit with when the subprocess is killed by a signal, I think we should at least try to follow [POSIX/SUS's rules for `$?` in shells](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_02):

> Internally, for purposes of deciding whether a command exits with a non-zero exit status, the shell shall recognize the entire status value retrieved for the command by the equivalent of the [wait()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html) function WEXITSTATUS macro (as defined in the System Interfaces volume of POSIX.1-2017). When reporting the exit status with the special parameter '?', the shell shall report the full eight bits of exit status available. The exit status of a command that terminated because it received a signal shall be reported as greater than 128.

yarikoptic:
good idea on reporting a signal name, let's do that @asmacdo .

As for exit -- we need to fix it, since ATM we wrap it around instead of adding abs one to 128:

```shell
❯ duct strace -t -f -o /tmp/`mdate`-nvidia-smi-strace-1.log nvidia-smi ; echo $?
[...]

PR #169, opened 2024-08-28
Modify exit code if cmd terminated by signal

asmacdo:
Fixes: https://github.com/con/duct/issues/148

Currently does not implement suggestion from @jwodder 

> If we're talking about what exit code duct should print when the subprocess is killed by a signal, I think "Exit Code: {n}" should be forgone in such circumstances and be replaced by something like "Killed by signal: {signal_name}".

This would be a good interface, but given that the "Summary" output at the end is user-customizable, I think it adds more complexity than its worth. IMO though, that feature may have outlived its usefulness as duct is getting more complex; @yarikoptic if we are ok with dropping `--summary-format`, I'd feel more comfortable adding logic to the summary.

### DIFF
--- a/test/test_execution.py
+++ b/test/test_execution.py
@@ -209,7 +209,7 @@ def test_signal_exit(temp_output_dir: str) -> None:
 
     thread = threading.Thread(target=runner)
     thread.start()
-    sleep(0.01)  # make sure the process is started
+    sleep(0.03)  # make sure the process is started
     ps_command = "ps aux | grep '[s]leep 60.74016230000801'"  # brackets to not match grep process
     ps_output = subprocess.check_output(ps_command, shell=True).decode()
     pid = int(ps_output.split()[1])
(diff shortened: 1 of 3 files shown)


test/test_execution.py
```diff
@@ -193,3 +196,29 @@ def test_execute_unknown_command(
...
+        args = Arguments.from_argv(
+            ["sleep", "60"],
+            output_prefix=temp_output_dir,
+        )
+        return execute(args)
+
+    thread = threading.Thread(target=runner)
+    thread.start()
+    sleep(0.01)  # make sure the process is started
+    ps_command = "ps aux | grep '[s]leep 60'"  # brackets to not match grep process
```
yarikoptic, line 213: fragile since could be some other process on accasion running some `sleep 60` etc.  since actual duration does not matter, make it larger and more likely to be "unique", e.g. 6903123

yarikoptic:
For this fix let's just fix the exit code value reporting . Let's discuss switch from format elsewhere (e.g. dedicated issue).

yarikoptic (approved):

PR #175, opened 2024-09-04
(Re)add etime and cmd into process stats

asmacdo:
Fixes https://github.com/con/duct/issues/174

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -279,11 +279,7 @@ class Sample:
         for pid in self.stats.keys() | other.stats.keys():
             if (mine := self.stats.get(pid)) is not None:
                 if (theirs := other.stats.get(pid)) is not None:
-                    try:
-                        output.add_pid(pid, mine.max(theirs))
-                    except Exception as e:
-                        lgr.critical(f"Pid: {pid} Mine: {mine} theirs: {theirs}")
-                        raise e
+                    output.add_pid(pid, mine.max(theirs))
                 else:
                     output.add_pid(pid, mine)
             else:
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -408,15 +408,17 @@ class Report:
                     "-s",
                     str(self.session_id),
                     "-o",
-                    "pid,pcpu,pmem,rss,vsz,etime,cmd",
+                    "pid,pcpu,pmem,rss,vsz,etime,stat,cmd",
                 ],
                 text=True,
             )
             for line in output.splitlines()[1:]:
                 if line:
-                    pid, pcpu, pmem, rss_kib, vsz_kib, etime, cmd = line.split(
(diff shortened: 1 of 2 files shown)


asmacdo:
Tests are failing on AssertionError   `assert self.cmd == other.cmd`

This shouldn't happen, could indicate a bigger issue-- process stat aggregation can only occur for the same process/cmd. I'll investigate whats going on with that.

yarikoptic:
And don't we have some integration test with known children processes ? (Some sleeps etc?)

yarikoptic:
and note that it doesn't fail consistently, so it is something flaky or incorrect assumption somewhere!

asmacdo:
2 failing tests were:
 - Test / test (ubuntu-latest, 3.12, py) (pull_request)
 - Test / test (ubuntu-latest, pypy-3.9, py) (pull_request)

Rerunning to see if its flake or version dependent.

asmacdo:
All checks passed. Rerunning CI for a sanity check since this has been flakey.

src/con_duct/__main__.py
```diff
@@ -126,14 +127,29 @@ class ProcessStats:
...
+    cmd: str
 
     def max(self, other: ProcessStats) -> ProcessStats:
+        cmd = self.cmd
+        if self.cmd != other.cmd:
+            lgr.debug(
+                f"cmd has changed. Previous measurement was {self.cmd}, now {other.cmd}."
+            )
+            # Brackets indicate that the kernel has substituted an abbreviation.
+            surrounded_by_brackets = r"\[.*?\]"
```
yarikoptic, line 140: why to overcomplicate (the `?`) and shouldn't it match at the beginning? otherwise it could be some argument etc.  And should be at least one character I think

```suggestion
            surrounded_by_brackets = r"^\[.+\]"
```

src/con_duct/__main__.py
```diff
@@ -126,14 +127,29 @@ class ProcessStats:
...
+            if re.search(surrounded_by_brackets, self.cmd):
+                lgr.debug(f"using {other.cmd}.")
+                cmd = other.cmd
```
yarikoptic, line 143: could we just keep that original one in?  might be interesting later for analysis etc.  It would be just something to remember that we need to go by PID (those IIRC shouldn't be readily reused so shouldn't overlap for different commands in consecutive records)

src/con_duct/__main__.py
asmacdo: +1

src/con_duct/__main__.py
asmacdo: We can't keep both with the current data structure-- both measurements are being aggregated into a single sample.

If the cmd has changed for real (with an exec or something?), we can still have different commands with the same pid in subsequent records. We just can't have 2 cmds in the same sample.

yarikoptic (approved):

PR #179, opened 2024-09-13
Fix operator precedence involving or and addition

asmacdo:
The addition expression was not evaluated when first operand is truthy. ie:
(1 or 0) + 1 == 2
1 or 0 + 1 == 1


yarikoptic:
coolio!  We should look into adding some linting -- I bet something could have warned us! (it is not our fault! ;))

Issue #172, opened 2024-09-03
Option to make <summary_format> still flexible for user 


asmacdo:
@yarikoptic since you seem attached to the summary_format  The user can set a `str`, (note, not an f string) https://github.com/con/duct/blob/main/src/con_duct/__main__.py#L31-L46

Since the user-provided value is a regular string, we cannot use fancing f-string logic.

Instead we could add the logic to what I'm calling the  "human readable preserializer" https://github.com/con/duct/blob/main/src/con_duct/__main__.py#L472-L476. That serializer is currently only responsible for replacing "unknown" string when the value is None, but it could be extended to do any other logic that would be necessary, ie rounding, and possibly even adding unit choices (ie Gb instead of bytes).

This would leave the user-customizable string to much simpler, and only allow the usage of pre-formatted values.

yarikoptic:
here is the format helper I pointed out in matrix yesterday:

> FWIW here is the formatter I was talking about I did for datalad probably 10 years ago by now: https://github.com/datalad/datalad-deprecated/blob/master/datalad_deprecated/ls.py#L369 . And then here how those formatters were used: https://github.com/datalad/datalad-deprecated/blob/master/datalad_deprecated/ls.py#L484 . So it seems we could easily allow for some !S for humanize.naturalsize wrapping around that value. and the construct for unknown I already gave you. Moreover there for percents we could even color similarly to how dstat does , ie. making them red if getting close to 100% or above.
> Do look at that code and see how ansi_colors where used etc - might come handy

I think it provides a very nice helper also to output "unknown" for `None` etc while concentrating all that formatting logic in the formatter instead of diluting through the code like proposed in

- #181

Issue #171, opened 2024-08-29
Do format % for CPU in summary sensibly/uniformly


yarikoptic:
I thought we had an issue for that, so feel free to close with a reference.

In a fresh https://github.com/con/duct/issues/170 I saw

```
    CPU Peak Usage: 12.2%
    Average CPU Usage: 6.457543859649121%
```
which should really be

```
    CPU Peak Usage: 12.2%
    Average CPU Usage: 6.4%
```
or alike -- there is really no value in those gory details for this one.

asmacdo:
@yarikoptic we have had some similar ones in the past. It would be easy to round all of the values in the "human readable" serializer. Currently, we do the rounding in the format string, which I had hoped would allow users to override the rounding. But the problem is that those strings can't handle non-float type errors.

```python
EXECUTION_SUMMARY_FORMAT_DEFAULT = (
    "Wall Clock Time: {wall_clock_time:.3f} sec\n"
    "Memory Average Usage (RSS): {average_rss:.3f} bytes\n"
)
```

But before I fix this, I am advocating that we instead drop the --summary-format feature instead, which I posed ([in the exit-code issue](https://github.com/con/duct/pull/169))

IMO this feature isn't super useful since this information is written in machine readable format in the info.json, so if the user wants to do anything with it, it would be better to use the file directly rather than parsing stderr logs.

yarikoptic:
before we drop anything -- I expect that all numbers are stored numerically in the record so you should be able to add similar :.1f or alike. Isn't that the case?

asmacdo:
No, in the case that we don't get a sample some of the values are None (which we translate to "unknown")

yarikoptic:
then could be `{"unknown" if blah is None else "%.1f" % blah}` or alike.  you could probably even add a helper, like `format_maybe_undefined` and use `{format_maybe_undefined(".1!f", blah)}`  or alike.

asmacdo:
user-provided format strings are strings, not f-strings.
See also https://github.com/con/duct/issues/172

yarikoptic:
here is a cute solution chatgpt inspired

```python
class MaybeNone:
    def __init__(self, value):
        self.value = value

    def __format__(self, format_spec):
        if self.value is None:
            return "unknown"
        else:
            return format(self.value, format_spec)

# Usage
unknown = MaybeNone(None)
pi = MaybeNone(3.14159)

print("{:.2f}".format(unknown))  # Output: unknown
[...]

PR #181, opened 2024-09-20
Round before formatting

asmacdo:
If we don't do the rounding prior to formatting, the summary_format (being a format string, not an f-string) is not smart enough to handle floats OR None (or str 'unknown'). So instead we need to handle that first.

Fixes: https://github.com/con/duct/issues/172
Fixes: https://github.com/con/duct/issues/171

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ options:
                         Output template to use when printing the summary
                         following execution. (default: Exit Code: {exit_code}
                         Command: {command} Log files location: {logs_prefix}
-                        Wall Clock Time: {wall_clock_time:.3f} sec Memory Peak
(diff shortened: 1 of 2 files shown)


yarikoptic:
Frankly I don't like this as it spreads aspects of formatting now from format specification into the code. as for "unknown" I already gave a nice solution to again allow it to be `float` or `None` (not sure some ad-hoc string) and then being nicely formatted: then you keep value nicely pythonic Optional[float] and format it the way you like it.   that is my 1c. and I gave already a dollar of those I think so far.

asmacdo:
Closing in favor of https://github.com/con/duct/pull/183

Issue #187, opened 2024-09-27
Add a test which `cp`s `__main__.py` as `duct` and executes basic test/run


yarikoptic:
While traveling I recalled why I wanted to keep `duct` (but not necessarily `duct-suite`) in a single file: to ease deployment/use as part of larger "runners on remote location". E.g. to replace "built-in" functionality within `reproman` from which we took the idea of capturing stdout/stderr.  For that we need a basic Python (no depends) script easy to `scp` and run.

For the same reason it should get shebang `!#/usr/bin/env python3`

PR #191, opened 2024-09-30
Move all logic into single file

asmacdo:
Should allow direct execution of file without need to pip install.

Fixes https://github.com/con/duct/issues/187


yarikoptic:
coolio, thanks!

PR #196, opened 2024-10-03
BF: Do not rely on having sources under ./src and __main__.py to be executable

yarikoptic:
Bad assumption since assumes running tests from the route of source distribution, that the `__main__.py` is executable and that sources are right nearby.  Pretty much all of those assumptions are violated during testing of built python package during debian packaging.

No immediate release necessary -- I will just pickup this patch for now

### DIFF
--- a/test/test_execution.py
+++ b/test/test_execution.py
@@ -4,10 +4,12 @@ import os
 from pathlib import Path
 import signal
 import subprocess
+import sys
 import threading
 from time import sleep, time
 import pytest
 from utils import assert_files
+import con_duct.__main__ as __main__
 from con_duct.__main__ import SUFFIXES, Arguments, Outputs, execute
 
 TEST_SCRIPT = str(Path(__file__).with_name("data") / "test_script.py")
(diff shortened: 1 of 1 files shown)


test/test_execution.py
```diff
@@ -227,6 +229,6 @@ def runner() -> int:
 
 
 def test_duct_as_executable(temp_output_dir: str) -> None:
-    ps_command = f"./src/con_duct/__main__.py -p {temp_output_dir} sleep 0.01"
+    ps_command = f"{sys.executable} {__main__.__file__} -p {temp_output_dir} sleep 0.01"
```
asmacdo, line 232: Isn't this the equivalent of 'python thefile.py'?

This makes sense to use the same version of python, but IIUC it doesn't prove the direct executability of the file.

yarikoptic: > Isn't this the equivalent of 'python thefile.py'?

yes

> This makes sense to use the same version of python, but IIUC it doesn't prove the direct executability of the file.

you are correct if by "executability" you mean an executable permissions/use of shebang.  The problem is that AFAIK that executable permission does not get copied while building/deploying `con-duct` so I had to specify the interpreter (`sys.executable`) directly to get it running.  The main point of the test though IMHO to verify that the script is "self sufficient" and does not rely on imports (ideally we should even copy it outside somewhere and run from there... later)

asmacdo (approved):

PR #198, opened 2024-10-08
Add `con-duct plot` with matplotlib backend

### LINKED ISSUE #106: Create helper script (con-duct plot?) to generate plots of memory/cpu through time
Might complement #105 nicely

I think with that we should initiate a helper "entry script" we could call `duct-suite` (or anything else) which would have support commands like this one to plot based on the stats collected for a process.  Could probably be by default overlap of memory and CPU consumption with left "y" axis to have CPU and right - memory.  Example for two graph plot with different scales: https://matplotlib.org/stable/gallery/subplots_axes_and_figures/two_scales.html 

Then we can define extra-depends for the `duct[suite]` to use matplotlib etc as desired for the `duct-suite`.

Note that might want to code right away so we might later provide different plotting backends (may be for inclusion online etc)

asmacdo:
Fixes #106

- [x] Add tests
~- [ ] Remove fixture usage.json and info.json files (replace with smaller files for tests)~

```
con-duct plot --help
usage: con-duct <command> [options] plot [-h] [--output OUTPUT] file_path

positional arguments:
  file_path        duct-produced usage.json file plot.

options:
  -h, --help       show this help message and exit
  --output OUTPUT  Output path for the png file. If not passed, show the file but do not save.
```

Example usage (with temporary files provided here):

```
$ con-duct plot --output usage test/data/mriqc-example/usage.json

Successfully rendered input file: test/data/mriqc-example/usage.json to output usage.png
```

usage


asmacdo:
pypy 3.8 is failing due to some import stuff in numpy. I think we can ignore this since we are dropping 3.8 which is EOL.
https://github.com/con/duct/pull/199

test/data/mriqc-example/usage.json
```diff
@@ -0,0 +1,6 @@
+{"timestamp": "2024-10-08T14:37:00.831368-04:00", "num_samples": 1, "processes": {"2703236": {"pcpu": 0.0, "pmem": 0.0, "rss": 40456192, "vsz": 1603104768, "timestamp": "2024-10-08T14:37:00.831368-04:00", "etime": "00:00", "stat": {"Ssl": 1}, "cmd": "singularity run --contain --bind /home/asmacdo/devel/sandbox/mriqc-sanity/sourcedata:/data:ro --bind /home/asmacdo/devel/sandbox/mriqc-sanity/long-run-4:/out --bind /home/asmacdo/devel/sandbox/mriqc-sanity/long-run-4/workdir:/workdir docker://nipreps/mriqc:24.0.2 /data /out participant --participant-label 02 -w /workdir --no-sub"}}, "totals": {"pmem": 0.0, "pcpu": 0.0, "rss": 40456192, "vsz": 1603104768}, "averages": {"rss": 40456192, "vsz": 1603104768, "pmem": 0.0, "pcpu": 0.0, "num_samples": 1}}
```
asmacdo, line 1: I had planned to replace this with a stripped down json, but on second thought, this isn't really that big-- maybe its best to pass a legit file with real data in it.

src/con_duct/suite/main.py
```diff
@@ -32,7 +13,7 @@ def execute(args: argparse.Namespace) -> int:
     return result
 
 
-def main() -> None:
+def main() -> None:  # pragma: no cover
```
yarikoptic, line 16: why "no cover"?   can make it

```suggestion
def main(args=None) -> None:  # pragma: no cover
```

and pass `args` below into `parse_args` and thus test parsing or at least e.g. `--help` to produce desired output etc

see e.g. smth similar in https://github.com/datalad/datalad/blob/HEAD/datalad/cli/main.py#L61 and then https://github.com/datalad/datalad/blob/HEAD/datalad/cli/tests/test_main.py#L113 and there on

asmacdo: I originally did this because we cant naively use the `Arguments` class like we do in main since the suite has multiple variants of potential args depending on the subparsers.

Ive added some basic tests but probably some room for improvement tbh.

yarikoptic:
please fix up types annotations, and let's get it merged to potentially improve later

yarikoptic:
ping on above - let's do fix up of type annotations (easy) and merge as is for now

PR #199, opened 2024-10-09
Drop Python 3.8, which is EOL 

asmacdo:
Happened October 7 https://devguide.python.org/versions/


yarikoptic:
is there any feature of 3.9 you are eager to add/start using? (pypy testing seems also error out on 3.9...)

yarikoptic:
spotted
- https://github.com/con/duct/pull/198#issuecomment-2402513929
so let's indeed just let it RiP

Issue #188, opened 2024-09-27
Remove `num_samples` and `num_reports` from default `summary_format`


asmacdo:
I found it to be a useful sanity check as a developer, but its probably not doing much for the user. 

Lets keep the variable so it could be used, but remove it from the default.

PR #200, opened 2024-10-09
Rm num_samples & num_reports from summary_format

asmacdo:
These variables are still available for users who provide a custom summary_format

Fixes https://github.com/con/duct/issues/188


asmacdo:
@yarikoptic Its been a bit since last release. Lets go ahead and cut one. Ready for review and merge if you like.

Issue #157, opened 2024-08-21
Test Python 3.13 starting October 1 2024


asmacdo:
[3.13.0rc1](https://docs.python.org/3.13/index.html)

PR #202, opened 2024-10-09
Add testing for Python 3.13

asmacdo:
Fixes https://github.com/con/duct/issues/157


test/test_execution.py
```diff
@@ -213,7 +213,7 @@ def runner() -> int:
 
     thread = threading.Thread(target=runner)
     thread.start()
-    sleep(0.03)  # make sure the process is started
+    sleep(0.1)  # make sure the process is started
```
asmacdo, line 216: Previously this error was caused by not waiting log enough. but now im skeptical (it could be another reason ps is returning a nonzero)
```
             if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command 'ps aux | grep '[s]leep 60.74016230000801'' returned non-zero exit status 1.
/opt/hostedtoolcache/Python/3.13.0/x64/lib/python3.13/subprocess.py:577: CalledProcessError
```

yarikoptic: what was stderr while running it?

is there `shell=True`?

yarikoptic: please try tmate action so you could login into github action environment, and try it 
"first hand"

asmacdo: Yes `shell=True`

With `tmate` action I was able to ssh into the action machine.

```
      - name: Setup tmate session
        uses: mxschmitt/action-tmate@v3
```

But ... `tox -e py -- -vv --cov-report=xml` passed?!  I'm not sure what I'm missing here, but this is confusing!

yarikoptic:
since also desired for conda IIRC, let's merge/release.  So merge whenever satisfied (I see some TMP, not sure if still in the queue)

PR #204, opened 2024-10-13
Explicitly mention con-duct command in the summary

asmacdo:


asmacdo:
Using a venv would work if we activated in every step, but the idiomatic way is just to use the setup-python action.

asmacdo:
@yarikoptic I want to cut this release because theres a problem with the conda build that looks related to the debian build issue you hit that has been fixed but not released.

```
=================================== FAILURES ===================================
___________________________ test_duct_as_executable ____________________________

temp_output_dir = '/tmp/pytest-of-conda/pytest-0/test_duct_as_executable0/'

        ps_command = f"./src/con_duct/__main__.py -p {temp_output_dir} sleep 0.01"
        # Assert does not raise
>       subprocess.check_output(ps_command, shell=True).decode()
```

yarikoptic:
sure, release then!

PR #216, opened 2024-10-29
Add released auto plugin to mark issues with releases where they were fixed

yarikoptic:


asmacdo:
Nice to see that it worked!

PR #226, opened 2024-11-15
Test abandoning parent

### LINKED ISSUE #44: Test with abandoning_parent
One of the reasons duct was implemented was to poll processes where the parent abandons the child process. 

Example test script: https://github.com/con/duct/blob/af67030a40fef7378418d5fa626d6baee7523436/abandoning_parent.sh

asmacdo:
~WIP: Passes, but ONLY when sleep is 0.1, longer periods result in too many PIDs??~

Fixes #44

### DIFF
--- a/test/test_e2e.py
+++ b/test/test_e2e.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 import subprocess
+import pytest
 
 ABANDONING_PARENT = str(Path(__file__).with_name("data") / "abandoning_parent.sh")
 
@@ -11,9 +12,9 @@ def test_sanity(temp_output_dir: str) -> None:
     subprocess.check_output(command, shell=True)
 
 
-def test_abandoning_parent(temp_output_dir: str) -> None:
+@pytest.mark.parametrize("num_children", [3, 5, 10, 20])
+def test_abandoning_parent(temp_output_dir: str, num_children: int) -> None:
     duct_prefix = f"{temp_output_dir}log_"
-    num_children = 3
     command = f"duct -p {duct_prefix} {ABANDONING_PARENT} {num_children} sleep 0.1"
     subprocess.check_output(command, shell=True)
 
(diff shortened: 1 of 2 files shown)


asmacdo:
pypy test failure is not related and will be fixed by https://github.com/con/duct/pull/233

test/test_e2e.py
```diff
@@ -0,0 +1,30 @@
...
+    command = f"duct --s-i 0.001 --r-i 0.01 -p {duct_prefix} {ABANDONING_PARENT} {num_children} sleep 0.35"
+    subprocess.check_output(command, shell=True)
+
+    with open(f"{duct_prefix}usage.json") as usage_file:
+        all_samples = [json.loads(line) for line in usage_file]
+
+    all_processes = {}
+    for sample in all_samples:
+        for pid, proc in sample["processes"].items():
+            all_processes[pid] = proc
```
yarikoptic, line 27: you don't care about values there, do you? then why bother, and not to have just a set... and anyways IMHO easier to read if just a comprehension like

```suggestion
    all_pids = {
        pid
        for pid in sample["processes"]
        for sample in all_samples
    }
```
but I bet there is even more succinct/clearer pattern with some itertools, ... may be later....

asmacdo: >  even more succinct/clearer pattern with some itertools

was pretty simple, thats what i did

test/test_e2e.py
```diff
@@ -0,0 +1,31 @@
...
+    command = f"duct --s-i 0.001 --r-i 0.01 -p {duct_prefix} {ABANDONING_PARENT} {num_children} sleep 0.35"
+    subprocess.check_output(command, shell=True)
+
+    with open(f"{duct_prefix}usage.json") as usage_file:
+        all_samples = [json.loads(line) for line in usage_file]
+
+    all_pids = {
+        pid
+        for pid in chain.from_iterable(sample["processes"] for sample in all_samples)
+    }
```
yarikoptic, line 28:
```suggestion
    all_pids = set(chain.from_iterable(sample["processes"] for sample in all_samples))
```
?

yarikoptic: someone really obnoxious could have even made it smth like ( I didn't test, but should work)

```suggestion
    all_pids = set(chain.from_iterable(map(operator.attrgetter("processes"), all_samples)))
```
but that is pushing it close/over the sensible limit in aim for de-duplication ;-)

asmacdo: I went with option 1 for readability, but the attrgetter with map is cool :)

PR #227, opened 2024-11-22
Add --fail-time option and by default remove all outputs if command fails fast

yarikoptic:
The most annoying UX for me to use duct "by default" is the fact that while preparing target command invocation, I can make mistakes.  And then duct keeps generating those reports which are inconsueqential and need cleaning (removal) etc. And such reports have no value really.  Hence, by default, I decided that it is not worth keeping them.  If it is desired, user can use this added option to set --fail-time to 0.  But also it might be that command takes awhile to start to only to fail, and for such cases users could raise it also higher and also have logs cleared for any failed command (set to -1).


jwodder (approved):

yarikoptic:
It is all green now.
I need to get it and a release so I could continue using it.  @asmacdo , feel welcome to leave comments or send a follow up PR if anything to be tuned up.

PR #232, opened 2025-01-14
DO NOT MERGE just running tests as-is

asmacdo:


asmacdo:
https://github.com/con/duct/pull/233 fixes the issue in this test.

Issue #184, opened 2024-09-20
on Ctrl-C  -- kill child not itself


yarikoptic:
Currently

```
❯ duct  dd if=/dev/zero of=/dev/null
2024-09-20T10:26:35-0400 [INFO    ] con-duct: duct is executing 'dd if=/dev/zero of=/dev/null'...
2024-09-20T10:26:35-0400 [INFO    ] con-duct: Log files will be written to .duct/logs/2024.09.20T10.26.35-184161_
^C^C^CTraceback (most recent call last):
  File "/usr/lib/python3.12/subprocess.py", line 1264, in wait
    return self._wait(timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/subprocess.py", line 2053, in _wait
    (pid, sts) = self._try_wait(0)
                 ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/subprocess.py", line 2011, in _try_wait
    (pid, sts) = os.waitpid(self.pid, wait_flags)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/yoh/proj/CON/duct/venvs/dev3.12/bin/duct", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/yoh/proj/CON/duct/src/con_duct/__main__.py", line 883, in main
    sys.exit(execute(args))
             ^^^^^^^^^^^^^
  File "/home/yoh/proj/CON/duct/src/con_duct/__main__.py", line 972, in execute
    process.wait()
  File "/usr/lib/python3.12/subprocess.py", line 1277, in wait
    self._wait(timeout=sigint_timeout)
  File "/usr/lib/python3.12/subprocess.py", line 2047, in _wait
    time.sleep(delay)
KeyboardInterrupt
^CException ignored in: <module 'threading' from '/usr/lib/python3.12/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.12/threading.py", line 1624, in _shutdown
    lock.acquire()
KeyboardInterrupt: 
```

so we are pretty much killing ourselves and do not provide summary etc.  I think we should

[...]

PR #235, opened 2025-01-22
Remove log files if interrupted via Ctrl-C

yarikoptic:
This is not how it should be!  it would depend on when it is interrupted and what was the actual process return code.  Immediate usecase is my impatience with current implementation still waiting whenever command such as  "ls" is done already.

Then when I interrupt with Ctrl-C we should still dump all the results and proceed to treat on how we would treat if command failed or not.  It would also relate to

- https://github.com/con/duct/issues/184

so if Ctrl-C the child process -- it is non-0 exit, and we would treat as "failed run".

So we should add it more smartly than what I did here

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -1021,93 +1021,102 @@ def execute(args: Arguments) -> int:
     files_to_close.append(report.usage_file)
 
     report.start_time = time.time()
-    try:
-        report.process = process = subprocess.Popen(
-            [str(args.command)] + args.command_args,
-            stdout=stdout_file,
-            stderr=stderr_file,
-            start_new_session=True,
-        )
-    except FileNotFoundError:
-        # We failed to execute due to file not found in PATH
-        # We should remove log etc files since they are 0-sized
-        # degenerates etc
+    try:  # Ctrl-C handling
+        try:
+            report.process = process = subprocess.Popen(
+                [str(args.command)] + args.command_args,
+                stdout=stdout_file,
+                stderr=stderr_file,
+                start_new_session=True,
+            )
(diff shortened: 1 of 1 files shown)


asmacdo:
@yarikoptic After fixed behavior from SIGINT, and the inclusion of `--fail-time`  I think we might have everything you were looking for.  Closing for now, feel free to reopen if we are still missing something.

```
$ duct sleep 3
2025-05-01T10:15:12-0500 [INFO    ] con-duct: duct is executing 'sleep 3'...
2025-05-01T10:15:12-0500 [INFO    ] con-duct: Log files will be written to .duct/logs/2025.05.01T10.15.12-1123278_
^C2025-05-01T10:15:13-0500 [INFO    ] con-duct: Received SIGINT, passing to command
2025-05-01T10:15:13-0500 [INFO    ] con-duct: Removing log files since command failed in less than 3.0 seconds.
```
```
$ duct --fail-time -1 sleep 10
2025-05-01T10:17:38-0500 [INFO    ] con-duct: duct is executing 'sleep 10'...
2025-05-01T10:17:38-0500 [INFO    ] con-duct: Log files will be written to .duct/logs/2025.05.01T10.17.38-1124219_
^C2025-05-01T10:17:44-0500 [INFO    ] con-duct: Received SIGINT, passing to command
2025-05-01T10:17:44-0500 [INFO    ] con-duct: Removing log files since command failed.
```

yarikoptic:
so, we would still keep logs if Ctrl-C'ed after `--fail-time` (default to 3 seconds)... So even though we have not addressed this issue to the letter, I think we are indeed ok  for now.

Issue #239, opened 2025-02-07
ls: add option to filter entries -e|-eval-filter PYTHON_EXPR


yarikoptic:
immediate usecase is inspired by use of `duct` within `datalad run` leading to overall run not committing since on of the commands did fail.

listing all is tedious and error prone.  Immediate desire is to limit by exit_code within record.  Workaround is to use `grep`

```
(3) yoh@typhon:/data/repronim/OpenNeuroDerivatives-NIDM$ grep -l '"exit_code": 1' .duct/logs/2025.*_info.json | xargs con-duct ls
PREFIX                                  COMMAND                              EXIT_CODE WALL_CLOCK_TIME PEAK_RSS
.duct/logs/2025.02.06T17.54.15-1500544_ code/do-segstats2nidm-perds ds001497 1         18.734 sec      404.3 MB
.duct/logs/2025.02.06T19.05.40-1871208_ code/do-segstats2nidm-perds ds003242 1         1.976 sec       33.4 MB 
```

I can also imagine desires to list only the runs for specific command  (regex), too long or too short tasks (`><` limit), and the same for memory. So a solution should not be really "exit_code" specific but rather allow for expressions.  In principle it could be a python expression to be evaluated on each record after providing `execution_summary`  as `locals`, e.g. here `--eval-filter 'exit_code==1'` ...

So -- implementation is trivial and here is an example . note that I added .copy() since otherwise eval seems could change that record and it would be not desirable

```
In [1]: import json

In [2]: import re

In [3]: j = json.load(open('.duct/logs/2025.02.06T19.05.40-1871208_info.json'))

In [4]: eval("exit_code==1 and command.startswith('code/')", j['execution_summary'].copy(), dict(re=re))
Out[4]: True

In [5]: eval("exit_code==1 and bool(re.search('code/', command))", j['execution_summary'].copy(), dict(re=re))
Out[5]: True
```

NB. another feature I immediately thought of is to add some lazy loading of actual log files... may be there could be some `stderr` and alike added to that record to pass, so if accessed, it would load the corresponding file and provide lazy-loaded content under their `.content` so interested parties would do smth like `"secret" in stderr.content`, or may be just expose them as `Path()` object, so people could explicitly do `stderr.read_text()` if desired... no immediate use case for it yet though.

PR #241, opened 2025-02-07
Add --eval-filter

asmacdo:
Fixes: https://github.com/con/duct/issues/239

- [x] Add tests
- [x] Investigate is _flatten_dict copy sufficient, or do we need deep_copy

`_flatten_dict` flattens the info dict, so we can eval-filter even parts that are not in execution_summary. Additionally, it creates a copy, so no need to .copy()

### DIFF
--- a/src/con_duct/suite/ls.py
+++ b/src/con_duct/suite/ls.py
@@ -55,7 +55,9 @@ LS_FIELD_CHOICES: List[str] = (
 MINIMUM_SCHEMA_VERSION: str = "0.2.0"
 
 
-def load_duct_runs(info_files: List[str], eval_filter: str) -> List[Dict[str, Any]]:
+def load_duct_runs(
+    info_files: List[str], eval_filter: Optional[str] = None
+) -> List[Dict[str, Any]]:
     loaded: List[Dict[str, Any]] = []
     for info_file in info_files:
         with open(info_file) as file:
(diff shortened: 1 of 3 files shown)


src/con_duct/suite/main.py
```diff
@@ -87,6 +87,10 @@ def main(argv: Optional[List[str]] = None) -> None:
         help="Path to duct report files, only `info.json` would be considered. "
         "If not provided, the program will glob for files that match DUCT_OUTPUT_PREFIX.",
     )
+    parser_ls.add_argument(
+        "-e",
+        "--eval-filter",
```
yarikoptic, line 92: doc?  how user would know what fields are available ? (in particular due to flatten_dict use)

src/con_duct/suite/ls.py
```diff
@@ -54,22 +55,29 @@
...
                     # TODO lower log level once --log-level is respected
                     lgr.warning(
                         f"Skipping {this['prefix']}, schema version {this['schema_version']} "
                         f"is below minimum schema version {MINIMUM_SCHEMA_VERSION}."
                     )
+                    continue
+                if eval_filter is not None and not eval(
+                    eval_filter, _flatten_dict(this), dict(re=re)
+                ):
+                    continue
```
yarikoptic, line 78:
```suggestion
                if eval_filter is not None and not (eval_results := eval(
                    eval_filter, _flatten_dict(this), dict(re=re)
                )):
                    lgr.debug("Filtering out %s due to filter results matching: %s", this, eval_results)
                    continue
```

PR #243, opened 2025-02-19
Add test: ls field list should contain all info.json fields

asmacdo:
This is built on https://github.com/con/duct/pull/241, originally to determine the list of fields that are filterable, but that is a secondary bonus, so I've separated this work into a separate PR. 

By running duct, and flattening the output info.json we can guarantee that the info.json field list has not changed. When it does, this test will fail, and can help us to remember to bump the schema version.

Additionally, this tests that LS_FIELD_CHOICES is accurate and contains all the field choices (currently we are missing a couple)

- [x] Add missing fields to LS_FIELD_CHOICES


yarikoptic:
per conversation -- no need for overcomplicating, just make a test to compare to a list of specific values (being sorted)

yarikoptic:
anything more to be done on this or out of draft? (please check other PRs also on that)

Issue #245, opened 2025-02-19
Make yaml optional dependecy


yarikoptic:
used for `-f yaml`

PR #248, opened 2025-02-21
bf: yaml should be optional

asmacdo:
Fixes: https://github.com/con/duct/issues/245


asmacdo:
Rebase after https://github.com/con/duct/pull/249

PR #251, opened 2025-02-27
Use versioningit for --version

### LINKED ISSUE #237: Show commit for --version when not a released version.
Current main 

```
duct --version
duct 0.8.0
```

Would be nice to be 
`duct 0.8.0-12345678`

asmacdo:
If a released version is installed, there should be no change, the git tag should be returned. If a local version is installed, then the version should include git hash.

Fixes #237

```sh
$ duct --version
duct 0.10.1.post11+g965a136
```

And, supposedly after the next release it will just be duct 0.11.0.

I am marking this as a draft because I want to double check how this will affect release workflow

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,0 @@
-[bumpversion]
-current_version = 0.11.0
-commit = True
-message = [skip ci] Bump version: {current_version} → {new_version}
-tag = False
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,5 +3,3 @@ current_version = 0.11.0
 commit = True
 message = [skip ci] Bump version: {current_version} → {new_version}
 tag = False
-
-[bumpversion:file:src/con_duct/__main__.py]
--- a/src/con_duct/__init__.py
+++ b/src/con_duct/__init__.py
@@ -1,0 +1,1 @@
+from ._version import __version__  # noqa
--- a/src/con_duct/__init__.py
+++ b/src/con_duct/__init__.py
@@ -1,1 +1,3 @@
-from ._version import __version__  # noqa
+from importlib.metadata import version
+
+__version__ = version("con-duct")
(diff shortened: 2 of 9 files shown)


yarikoptic:
@asmacdo has a concern if auto's "bump to version" would  work out correctly after this change. @jwodder WDYT? please review

src/con_duct/__init__.py
```diff
@@ -0,0 +1 @@
+from ._version import __version__  # noqa
```
jwodder, line 1: Why not do `__version__ = importlib.metadata.version("con-duct")` like in `__main__.py`?  Using two different methods for getting the version at runtime seems silly.

.bumpversion.cfg
```diff
@@ -3,5 +3,3 @@ current_version = 0.10.1
 commit = True
 message = [skip ci] Bump version: {current_version} → {new_version}
 tag = False
-
-[bumpversion:file:src/con_duct/__main__.py]
```
jwodder, line 7: This whole file should be deleted if you're switching to `versioningit`, and the "afterChangelog" line in `.autorc` should be deleted.

jwodder (changes requested):

yarikoptic:
how far is this one from completion/being "open" for review?

asmacdo:
@yarikoptic it could be reviewed now-- i am trying to run `auto shipit --dry-run` locally to test how this would work, but that shouldnt block review

asmacdo:
AFAICT, `auto shipit --dry-run` (or canary) isn't really going to test what we need. Running complains `⚠  warning   None of the plugins that you are using implement the `canary` command!` 

 I don't see a way to verify ahead of merge that this will work. I think we just merge this and rely on `auto` to do its thing, and `versioningit` to do its thing.

asmacdo:
I added a release tag to this-- so if something does go wrong, we will at least know quickly.

PR #257, opened 2025-03-07
When using pyout, use pyout color

asmacdo:
image

##TODO
- [x] We need a way to disable colors for pyout
- [x] Consolidate VALUE_TRANSFORMATION_MAP custom converters (see duplication below)

duplication:
```python
VALUE_TRANSFORMATION_MAP: Dict[str, str] = {                                                                                               
    "average_pcpu": "{value:.2f!N}%",                                                                                                      
    "average_pmem": "{value:.2f!N}%",                                                                                                      
    "average_rss": "{value!S}",                                                                                                            
    "average_vsz": "{value!S}",                                                                                                            
    "end_time": "{value:.2f!N}",                                                                                                           
    "exit_code": "{value!E}",                                                                                                              
    "memory_total": "{value!S}",                                                                                                           
    "peak_pcpu": "{value:.2f!N}%",                                                                                                         
    "peak_pmem": "{value:.2f!N}%",                                                                                                         
    "peak_rss": "{value!S}",                                                                                                               
    "peak_vsz": "{value!S}",                                                                                                               
    "start_time": "{value:.2f!N}",                                                                                                         
    "wall_clock_time": "{value:.3f} sec",                                                                                                  
}   
```

```python
    with pyout.Tabular(                                                                                                                    
        style=dict(                                                                                                                        
            header_=dict(bold=True, transform=str.upper),                                                                                  
            average_pcpu=N,                                                                                                                
            average_pmem=N,                                                                                                                
            average_rss=S,                                                                                                                 
            average_vsz=S,                                                                                                                 
            end_time=N,                                                                                                                    
            exit_code=E,                                                                                                                   
            memory_total=S,                                                                                                                
            peak_pcpu=N,                                                                                                                   
            peak_rss=S,                                                                                                                    
            peak_vsz=S,                                                                                                                    
            start_time=N,                                                                                                                  
        ),                                                                                                                                 
        mode="final",                                                                                                                      
    ) as table:                                                                                                                            
        for row in run_data_list:                                                                                                          
            table(row) 
   ```

### DIFF
--- a/src/con_duct/suite/ls.py
+++ b/src/con_duct/suite/ls.py
@@ -148,14 +148,46 @@ def _format_row(
     return transformed
 
 
-def pyout_ls(run_data_list: List[OrderedDict[str, Any]]) -> None:
+def pyout_ls(run_data_list: List[OrderedDict[str, Any]], enable_colors: bool) -> None:
     """Generate and print a tabular table using pyout."""
     if pyout is None:
         raise RuntimeError("pyout is required for this output format.")
 
+    color_styles = {
+        "E": dict(
+            color=dict(
+                re_lookup=[
+                    ["^0$", "green"],  # if exactly "0", then green
+                    [".*", "red"],  # anything else gets red
+                ]
+            )
+        ),
+        "N": dict(
+            color=dict(
+                re_lookup=[
+                    [f"^{SummaryFormatter.NONE}", "red"],  # if starts with NONE
+                    [".*", "green"],
+                ]
+            )
+        ),
+    }
+    # S is humansize, conversion done, coloring same as N
+    color_styles["S"] = color_styles["N"]
+
+    pattern = re.compile(r"!([A-Z])")
+    conversion_map = (
+        {
+            k: color_styles[match.group(1)]
+            for k, v in VALUE_TRANSFORMATION_MAP.items()
+            if (match := pattern.search(v))
+        }
+        if enable_colors
+        else {}
+    )
     with pyout.Tabular(
         style=dict(
             header_=dict(bold=True, transform=str.upper),
+            **conversion_map,
         ),
         mode="final",
     ) as table:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/con_duct/suite/ls.py [hunk 1]
```diff
@@ -153,9 +153,37 @@ def pyout_ls(run_data_list: List[OrderedDict[str, Any]]) -> None:
...
+            average_pmem=N,
+            average_rss=S,
+            average_vsz=S,
+            end_time=N,
+            exit_code=E,
+            memory_total=S,
+            peak_pcpu=N,
+            peak_rss=S,
+            peak_vsz=S,
+            start_time=N,
```
yarikoptic, line 186: per agreement -- replace with a little `**{}` comprehension over the other style to instruct which one to use. And may be for that replace individual variables with a `modifiers = {'S': ..., 'N': ...}` so you could just take the "modifier" character using regex from the original format.

src/con_duct/suite/ls.py
```diff
@@ -164,19 +192,22 @@ def pyout_ls(run_data_list: List[OrderedDict[str, Any]]) -> None:
...
         args.paths = [p for p in glob.glob(pattern)]
 
+    if args.format == "auto":
+        args.format = "summaries" if pyout is None else "pyout"
+
+    if args.format == "pyout":
+        formatter = SummaryFormatter(enable_colors=False)
+    else:
+        formatter = SummaryFormatter(enable_colors=args.colors)
+
```
yarikoptic, line 206:
```suggestion
        formatter = SummaryFormatter(enable_colors=False if args.format == "pyout" else args.colors)
```

src/con_duct/suite/ls.py
yarikoptic: and here keep it empty dict if no colors

jwodder (commented):
I do not know enough about pyout to review this.

Issue #252, opened 2025-02-27
CI Flake: Abandoning Parent test


asmacdo:
=================================== FAILURES ===================================
__________________________ test_abandoning_parent[1] ___________________________

temp_output_dir = '/tmp/pytest-of-runner/pytest-0/test_abandoning_parent_1_0/'
num_children = 1

    @pytest.mark.parametrize("num_children", [1, 2, 10])
    def test_abandoning_parent(temp_output_dir: str, num_children: int) -> None:
        duct_prefix = f"{temp_output_dir}log_"
        command = f"duct --s-i 0.001 --r-i 0.01 -p {duct_prefix} {ABANDONING_PARENT} {num_children} sleep 0.2"
        subprocess.check_output(command, shell=True)
    
        with open(f"{duct_prefix}usage.json") as usage_file:
            all_samples = [json.loads(line) for line in usage_file]
    
        all_pids = set(chain.from_iterable(sample["processes"] for sample in all_samples))
    
        # 1 for each child, 1 for pstree, 1 for parent
>       assert len(all_pids) == num_children + 2
E       AssertionError: assert 4 == (1 + 2)
E        +  where 4 = len({'2626', '2627', '2633', '2635'})

test/test_e2e.py:28: AssertionError
----------------------------- Captured stderr call -----------------------------
2025-02-27T19:58:39+0000 [INFO    ] con-duct: duct is executing '/home/runner/work/duct/duct/test/data/abandoning_parent.sh 1 sleep 0.2'...
2025-02-27T19:58:39+0000 [INFO    ] con-duct: Log files will be written to /tmp/pytest-of-runner/pytest-0/test_abandoning_parent_1_0/log_
2025-02-27T19:58:39+0000 [INFO    ] con-duct: Summary:
Exit Code: 0
Command: /home/runner/work/duct/duct/test/data/abandoning_parent.sh 1 sleep 0.2
Log files location: /tmp/pytest-of-runner/pytest-0/test_abandoning_parent_1_0/log_
Wall Clock Time: 0.254 sec
Memory Peak Usage (RSS): 7.3 MB
Memory Average Usage (RSS): 6.7 MB
Virtual Memory Peak Usage (VSZ): 20.4 MB
Virtual Memory Average Usage (VSZ): 18.6 MB
Memory Peak Percentage: 0.00%
Memory Average Percentage: 0.00%
CPU Peak Usage: 0.00%
Average CPU Usage: 0.00%

=========================== short test summary info ============================
[...]

PR #258, opened 2025-03-12
test: test various spawned children e2e cases

asmacdo:
This replaces the "abandoning_parent" tests.

One interesting point came up here though-- when there is a process that spawns other processes, **duct tracks abandoned processes, but does not wait for abandoned processes to finish**

Fixes https://github.com/con/duct/issues/252


test/data/spawn_children.sh
```diff
@@ -0,0 +1,27 @@
...
+for i in $(seq 1 $nchildren); do
+    case "$mode" in
+        subshell)
+            ( "$@" & ) ;;
+        nohup)
+            ( nohup "$@" & disown ) & ;;
+        setsid)
+            setsid "$@" & ;;
+        plain)
+            "$@" & ;;
```
yarikoptic, line 19: nice -- finally I see some disown for true abandonment ;)

test/test_e2e.py
```diff
@@ -5,24 +5,29 @@
...
     with open(f"{duct_prefix}usage.json") as usage_file:
         all_samples = [json.loads(line) for line in usage_file]
 
     all_pids = set(chain.from_iterable(sample["processes"] for sample in all_samples))
 
-    # 1 for each child, 1 for pstree, 1 for parent
-    assert len(all_pids) == num_children + 2
+    # Add 1 for the parent, add 1 for the "hold the door" process, see line 7 of the script
+    if mode == "setsid":
+        assert len(all_pids) == 2
```
yarikoptic, line 31: so we do not track/kill children correctly if underlying process (re)sets the sid, correct?

yarikoptic: I wonder if we need to indeed resort to the walking the tree of processes to figure them all out!  @vmdocua did similar in reprostim IIRC

asmacdo: Correct that's a fundamental limitation

yarikoptic (approved):

asmacdo:
This didn't fix the flake, which I suspect is because the "hold-the-door" process sleeps exactly as long as the children, so the parent process can exit slightly before the children. In our test, that means that there might sometimes be a previous child showing up in the next run of the test. 

```
 FAILED test/test_e2e.py::test_spawn_children[1-plain] - AssertionError: assert 4 == (1 + 2)
 +  where 4 = len({'2362', '2365', '2356', '2361'})
======================== 1 failed, 217 passed in 26.22s ========================
py: exit 1 (26.86 seconds) /home/runner/work/duct/duct> pytest -vv --cov-report=xml test pid=2269
  py: FAIL code 1 (44.18=setup[17.32]+cmd[26.86] seconds)
  evaluation failed :( (46.46 seconds)
  ```
  
  The solution I've implemented is making sure the "hold-the-door" process sleeps 1.5 times as long as any child process.

test/test_e2e.py
asmacdo: https://github.com/con/duct/issues/259

asmacdo:
Phew-! Upon further inspection the flake was due to forking of the parent process-- specifically, 

`/bin/bash /home/runner/work/duct` with a stat of "D"-- which I assume is related to the stdout stderr files? There was also a zombie process entry for what had been  `/bin/bash /home/runner/work/duct.  What this demonstrates is that the parent process does doe some forking, but probably at such short intervals they are not always represented in a duct sample.

Either way-- lets not make this test have to account for every possible fork of the parent process. Instead, lets restrict the test count to the children, which is the actual thing we are testing! This makes the result a lot simpler, and does seem to have solved the flake-- passing all test envs 6x in a row.

PR #260, opened 2025-03-14
Pass Ctrl+c interrupt to executed command

asmacdo:
https://github.com/con/duct/issues/184

### DIFF
--- a/test/test_execution.py
+++ b/test/test_execution.py
@@ -282,8 +282,14 @@ def test_signal_kill(temp_output_dir: str, fail_time: float | None) -> None:
 
 
 def test_duct_as_executable(temp_output_dir: str) -> None:
-    ps_command = (
-        f"{sys.executable} {__main__.__file__} -p {temp_output_dir} -q sleep 0.01"
-    )
+    ps_command = [
+        sys.executable,
+        __main__.__file__,
+        "-p",
+        temp_output_dir,
+        "-q",
+        "sleep",
+        "0.01",
+    ]
     # Assert does not raise
-    subprocess.check_output(ps_command, shell=True).decode()
+    subprocess.check_output(ps_command, shell=False).decode()


asmacdo:
@yarikoptic summary since we discussed:
 
 I ran these tests with coverage disabled, and they passed. https://github.com/con/duct/pull/260/commits/78f5c53ee3e5a17b505a1045f77c1aad8717598c
 Then i re-enabled coverage and just increased the wait time. 
 
 I'm very surprised that increased wait time between signals would have any effect since SIGINT is basically ignored in the signal_ignorer.py.  Also 0.3 seconds (previous) really seems like it should have been enough? 
 
 I cant say that this is "solved" but I'd argue it is "good enough, we can come back to it if it causes flake"

asmacdo:
The other remaining mystery is why the handling is not showing up in coverage. 

We explicitly test that there is a SIGKILL, so I'm guessing this could have something to do with the internals of how coverage works with multiprocessing? https://github.com/con/duct/pull/260/files#diff-
b00f6c135016a6447359352933c8f84365d26d2c22db2bc833e0a57b387b1f8fR281

So im still curious, but probably not worth spending more time right now, I vote to merge as-is

test/test_execution.py
```diff
@@ -215,39 +220,70 @@ def runner() -> int:
...
+        args = Arguments.from_argv([script_path], output_prefix=temp_output_dir, **kws)
+        return execute(args)
+
+    wait_time = 0.8
+    proc = multiprocessing.Process(target=runner)
+    proc.start()
+    sleep(wait_time)
+    assert proc.pid is not None, "Process PID should not be None"  # for mypy
+    os.kill(proc.pid, signal.SIGINT)
+    sleep(wait_time)
```
yarikoptic, line 263: do we really need to wait that long too between our interruptions here or could may be proceed faster?

anything here verifies that `duct` actually tries to stop the underlying process?  may be the signal ignorer should not just sleep but do some "compute" as to switch from sleep into action, smth like

```python
t0 = time()
while time.time() - t0 < 10:
   sleep(0.01)
```
so it might response more promptly to attempts to be killed since would not just be sleeping all the time.

asmacdo: Thats a good idea, maybe that will make it faster to actually catch the interrupt?

There is some race condition weirdness with the signals when using both pypy and codecov.

asmacdo: One alternative to speed up the tests would be to run coverage only on one of cpython runs, then we could lower the wait times-- FWIW these passed even on pypy when cov was disabled.

asmacdo:
@yarikoptic I've left the commits for review purposes, lets not merge until squashed a bit. Marking as a draft to block merge, but ready for review.

test/test_execution.py
```diff
@@ -215,39 +220,70 @@ def runner() -> int:
...
-        assert exit_code == 128 + 15
+        command_exit_code = info_data["execution_summary"]["exit_code"]
+        # SIGKILL
+        assert command_exit_code == 128 + 9
 
 
 def test_duct_as_executable(temp_output_dir: str) -> None:
-    ps_command = f"{sys.executable} {__main__.__file__} -p {temp_output_dir} sleep 0.01"
+    ps_command = (
+        f"{sys.executable} {__main__.__file__} -p {temp_output_dir} -q sleep 0.01"
```
yarikoptic, line 286: since you are touching this, needs quotes around arguments or a `list` representation used so that if there is a space in any path - does not crash . Best would be a list form if possible here

yarikoptic: I don't have any further comments atm , IIRC it worked when I tried, so - let' proceed overall

PR #268, opened 2025-05-06
Test fix conda feedstock

asmacdo:
To be merged after https://github.com/con/duct/pull/267 so the changelog entries are separate.


asmacdo:
Marked as draft so this one isnt accidentally merged prior to #267

PR #269, opened 2025-05-08
Allow con-duct ls to function back to schema 0.2.0

### LINKED ISSUE #266: Provide upgrade migration for the schema when loading
Similarly to what we do in dandi-schema (and elsewhere, like pynwb etc) while loading json, check its schema_version and if below the version which added a new feature, just populate with `None`, e.g. currently smth like in https://github.com/dandi/dandi-schema/blob/HEAD/dandischema/metadata.py#L340

asmacdo:
Fixes #266

### DIFF
--- a/src/con_duct/suite/ls.py
+++ b/src/con_duct/suite/ls.py
@@ -103,9 +103,8 @@ def ensure_compliant_schema(info_dict):
     if parse_version(info_dict["schema_version"]) == parse_version(__schema_version__):
         return
     # working_directory added in 0.2.1
-    elif parse_version(info_dict["schema_version"]) < parse_version("0.2.1"):
+    if parse_version(info_dict["schema_version"]) < parse_version("0.2.1"):
         info_dict["execution_summary"]["working_directory"] = ""
-        return
 
 
 def process_run_data(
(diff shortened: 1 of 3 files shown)


src/con_duct/suite/ls.py
```diff
@@ -98,6 +99,15 @@ def load_duct_runs(
     return loaded
 
 
+def ensure_compliant_schema(info_dict):
+    if parse_version(info_dict["schema_version"]) == parse_version(__schema_version__):
+        return
+    # working_directory added in 0.2.1
+    elif parse_version(info_dict["schema_version"]) < parse_version("0.2.1"):
+        info_dict["execution_summary"]["working_directory"] = ""
+        return
```
yarikoptic, line 108: Most likely there would later be more of `if`s and not necessarily elif's , so might be worth right away to 

```suggestion
```

yarikoptic:
coolio, tests aren't happy though, need to enable that mergify and see if would be of assistance here...

asmacdo:
Rebased on mergify, typing should fail, lets see what it says!!

asmacdo:
oh well, mergify isnt used in typing, we can check it out on the next one :)

PR #270, opened 2025-05-09
Bolt on mergify support to get summaries from CI reported

yarikoptic:
Following loosely https://docs.mergify.com/ci-insights/pytest/ , and having defined the token org wide.

My hope for it to gain summaries of test fails across runs.


asmacdo:
Looks fine to me, lets merge and see what it does!

PR #271, opened 2025-05-09
bf: define pyci environment and there add pytest-mergify as dependency

yarikoptic:
Pre-installing does not affect environment tox would use for testing. To check: whether we end up with that original python version as before

Continuation to 
- #270 
where it was not really in effect. CI tox run should list mergify among plugins

```
plugins: cov-6.1.1, mergify-2025.4.10.1
```

but I do not want it to be added to default test env, to not always demand it


yarikoptic:
I am not sure yet what CI INSIGHTS would bring there for us, but at least seems to be operating now.

PR #275, opened 2025-05-29
auto push cleanup to PRs

asmacdo:
TODO:
- [x] use a bot to push the changes instead of the gh-actions default
- [ ] remove the intermediate changes (introduced to demonstrate that this action works)

### DIFF
--- a/.github/workflows/auto-fix-content.yaml
+++ b/.github/workflows/auto-fix-content.yaml
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout PR branch
         uses: actions/checkout@v4
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.GITMATE_GITHUB_TOKEN }}
           ref: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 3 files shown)


asmacdo:
@yarikoptic do you have a bot with gh credentials? The functionality here seems to be working correctly, but IIUC, since the commit is pushed by gh-actions itself, that doesn't trigger a new round of checks to prevent recursion, so to retrigger the checks I had to close and reopen the PR.

yarikoptic:
I think I do. But I am not following how it relates here

asmacdo:
If we use a bot with credentials that aren't the default GitHub action, then I think it's pushes will trigger CI to run the tests. As is, the tests don't run on the automatic follow up commit.

asmacdo:
There are some other dandi repos that use https://pre-commit.ci/ (I couldnt find it because AFAICT nothing is configured in the repos, but is a fully external service).

yarikoptic:
ok, we have `GITMATE_GITHUB_TOKEN` secret now, please use that secret.

asmacdo:
@yarikoptic Looks like its working. would you like to review this before I squash out the intermediate commits that were used to test this workflow?

Issue #276, opened 2025-06-13
Scheduled tests started to fail


yarikoptic:
![Image](https://github.com/user-attachments/assets/e47c58d1-ab65-4226-a760-99c650793f4d)

asmacdo:
This looks like flake, fortunately its always this same test. 


Run 1055 failed on python 3.11
```
________________________ test_spawn_children[1-setsid] _________________________
<snip>...
         if mode == "setsid":
>           assert len(all_child_pids) == 1
E           AssertionError: assert 2 == 1
E            +  where 2 = len({'2710', '2708'})
```

1054 failed on pypy 3.9

```
_________________________ test_spawn_children[1-nohup] _________________________
         else:
>           assert len(all_child_pids) == num_children + 1
E           assert 0 == (1 + 1)
E            +  where 0 = len(set())
```


1053 Failed on pypy 3.9 and pypy 3.10


pypy 3.9
```
________________________ test_spawn_children[2-setsid] _________________________
         if mode == "setsid":
>           assert len(all_child_pids) == 1
E           assert 0 == 1
E            +  where 0 = len(set())
```

pypy 3.10
```
________________________ test_spawn_children[10-setsid] ________________________
         if mode == "setsid":
>           assert len(all_child_pids) == 1
E           assert 0 == 1
E            +  where 0 = len(set())
```

For the ones where there are 0 child pids, we could just try increasing the sleep duration, currently set to 0.2.

But the test that had 2 child pids when there should have been 1... thats odd. Ill need to look into it more

PR #277, opened 2025-06-13
Reduce flake: increase test_spawn_children sleep dur and add retries

asmacdo:
Fixes: https://github.com/con/duct/issues/276

Since these tests always pass locally for me, I'm using this PR to try out some options for removing this flake.

### DIFF
--- a/test/test_e2e.py
+++ b/test/test_e2e.py
@@ -14,7 +14,9 @@ def test_sanity(temp_output_dir: str) -> None:
 
 
 @pytest.mark.parametrize("mode", ["plain", "subshell", "nohup", "setsid"])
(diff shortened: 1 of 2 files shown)


asmacdo:
RUN 1: passed

asmacdo:
RUN 2: passed

asmacdo:
Run 3: passed

I think we have a decent solution here to the immediate problem, but I think its going to be worth deeper investigation. If there are unavoidable edge cases, what are they? 

Under-reporting seems plausible. maybe if the system is bogged down, the sampling rate is slower than configured? 
Over-reporting I don't see how this is possible yet.

yarikoptic (approved):
let's see if it improves our quality of life

Issue #290, opened 2025-08-27
plot: does not show anything (ran without `--output`)


yarikoptic:
```shell
❯ con-duct plot .duct/logs/2025.08.27T10.43.14-157464_usage.json
/home/yoh/proj/HED/hed-python/.venv/lib/python3.11/site-packages/con_duct/suite/plot.py:59: UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown
  plt.show()
```

expectation: should visualize the plot .

should likely inspect what is the matplotlib backend ATM and if anything - provide user instructions on what to do next.  Note that if `--output` is specified, then it is expected to have non-interactive.

yarikoptic:
regular terminal.  forgot to list which environment, most likely a fresh `uv` with minimal set of dependencies installed. like `con-duct` and may be `matplotlib` and that is it. may be that is a reason.  ATM I fail to recall for which project that was to identify the uv venv to try at.

yarikoptic:
<details>
<summary>here is reproduction with all gory details</summary> 

```shell
❯ uv venv
Using CPython 3.11.12
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
❯ source .venv/bin/activate
❯ uv pip install con-duct matplotlib
Resolved 12 packages in 551ms
Prepared 1 package in 566ms
Installed 12 packages in 33ms
 + con-duct==0.15.0
 + contourpy==1.3.3
 + cycler==0.12.1
 + fonttools==4.59.2
 + kiwisolver==1.4.9
 + matplotlib==3.10.5
 + numpy==2.3.2
 + packaging==25.0
 + pillow==11.3.0
 + pyparsing==3.2.3
 + python-dateutil==2.9.0.post0
 + six==1.17.0
❯ con-duct ls | head
PREFIX                                 COMMAND                                                                                                      EXIT_CODE WALL_CLOCK_TIME PEAK_RSS
.duct/logs/2025.04.22T11.15.01-197335_ wget -S https://bossdb-open-data.s3.amazonaws.com/morgan2020/lgn/em/8_8_30/17920-18432_46592-47104_8272-8288 0         1.256 sec       11.1 MB 
.duct/logs/2025.04.22T11.15.28-197391_ wget -S https://bossdb-open-data.s3.amazonaws.com/morgan2020/lgn/em/8_8_30/17920-18432_46592-47104_8272-8288 0         1.590 sec       11.4 MB 
[...]

PR #293, opened 2025-08-28
Handle noninteractive matplotlib backends

asmacdo:
fixes https://github.com/con/duct/issues/290

I'd have called this a patch instead of a minor bump, but this will also drop support for matplotlib 3.9 since `matplotlib.get_backend` was introduced in 3.10. 

Unfortunately matplotlib 3.9 is not available with python 3.9, so I think we should hold this PR until 3.9 EOL in October 2025. 

- [ ] ~~Drop python 3.9 before merge~~

### DIFF
--- a/src/con_duct/suite/plot.py
+++ b/src/con_duct/suite/plot.py
@@ -4,11 +4,15 @@ import json
 
 
 def matplotlib_plot(args: argparse.Namespace) -> int:
-    import matplotlib
-    from matplotlib.backends import backend_registry
-    from matplotlib.backends.registry import BackendFilter
-    import matplotlib.pyplot as plt
-    import numpy as np
+    try:
+        import matplotlib
+        from matplotlib.backends import backend_registry
+        from matplotlib.backends.registry import BackendFilter
+        import matplotlib.pyplot as plt
+        import numpy as np
+    except ImportError as e:
+        print(f"con-duct plot missing required dependency: {e}")
+        return 1
 
     # Handle info.json files by reading the usage path from the file
     file_path = args.file_path
--- a/src/con_duct/suite/plot.py
+++ b/src/con_duct/suite/plot.py
@@ -6,7 +6,7 @@ import json
 def matplotlib_plot(args: argparse.Namespace) -> int:
     try:
         import matplotlib
-        from matplotlib.backends import backend_registry
+        from matplotlib.backends import backend_registry  # type: ignore[attr-defined]
         from matplotlib.backends.registry import BackendFilter
         import matplotlib.pyplot as plt
         import numpy as np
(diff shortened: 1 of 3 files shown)


yarikoptic:
Let's bring this little boy over the finish line @asmacdo

asmacdo:
There was a way to check backend before 3.10, so I've just fallen back to that. Hand tested, works fine.

src/con_duct/suite/plot.py
```diff
@@ -77,7 +84,32 @@ def matplotlib_plot(args: argparse.Namespace) -> int:
...
+                f"which is a not a known interactive backend."
+            )
+            print(
+                "Either set environment variable MPLBACKEND to an interactive backend or "
+                "use --output to save the plot to a file instead."
+            )
+            print(
+                "For more info: https://matplotlib.org/stable/users/explain/figure/backends.html"
+            )
+            return 1
```
yarikoptic, line 112: this is a CLI way (at best) to report... in Python libraries we should throw exceptions and use `logging` instead of prints unless the command intended to print, like `pp` or `ls`... it seems that it is the `plot` in the suite which abuses prints ATM even in the suite in main -- so we better do not breed even more of those:

```
❯ git grep '\<print\>' -- src/con_duct/suite
src/con_duct/suite/ls.py:    """Generate and print a tabular table using pyout."""
src/con_duct/suite/ls.py:                print(f"{col.replace('_', ' ').title()}: {value}")
src/con_duct/suite/ls.py:        print(json.dumps(output_rows))
src/con_duct/suite/ls.py:        print(json.dumps(output_rows, indent=2))
src/con_duct/suite/ls.py:        print(yaml.dump(plain_rows, default_flow_style=False))
src/con_duct/suite/main.py:    parser_pp = subparsers.add_parser("pp", help="Pretty print a JSON log.")
src/con_duct/suite/main.py:    parser_pp.add_argument("file_path", help="JSON file to pretty print.")
src/con_duct/suite/plot.py:            print(f"Error reading info file {args.file_path}: {e}")
src/con_duct/suite/plot.py:        print(f"File {file_path} was not found.")
src/con_duct/suite/plot.py:        print(f"File {file_path} contained invalid JSON.")
src/con_duct/suite/plot.py:        print(f"Usage file {file_path} is missing required field: {e}")
src/con_duct/suite/plot.py:        print(f"Usage file {file_path} contains invalid data format: {e}")
src/con_duct/suite/plot.py:        print(f"Error processing usage file {file_path}: {e}")
src/con_duct/suite/plot.py:        print(f"Successfully rendered input file: {file_path} to output {args.output}")
src/con_duct/suite/pprint_json.py:        print(f"File not found: {args.file_path}")
src/con_duct/suite/pprint_json.py:        print(f"Error decoding JSON: {e}")

```

Issue #299, opened 2025-09-05
plot: seems to derange into using relative path having being provided a full one


yarikoptic:
```shell
❯ uvx --with matplotlib con-duct plot /home/yoh/proj/datalad/datasets.datalad.org-gitlog-stats/.duct/logs/2025.03.30T19.37.05-4137075_info.json
ERROR:con_duct.suite.plot:File .duct/logs/2025.03.30T19.37.05-4137075_usage.json was not found.
❯ ls -ld /home/yoh/proj/datalad/datasets.datalad.org-gitlog-stats/.duct/logs/2025.03.30T19.37.05-4137075_info.json
-rw-rw-r-- 1 yoh yoh 1258 Mar 31 10:03 /home/yoh/proj/datalad/datasets.datalad.org-gitlog-stats/.duct/logs/2025.03.30T19.37.05-4137075_info.json
❯ ls -ld /home/yoh/proj/datalad/datasets.datalad.org-gitlog-stats/.duct/logs/2025.03.30T19.37.05-4137075_info.json
❯ uvx --with matplotlib con-duct --version
con-duct 0.16.0
```

so it is not taking _usage file relative to the _info after loading from the _info.json

PR #301, opened 2025-09-09
zfix: plot usage with info.json abs path

asmacdo:
Fixes https://github.com/con/duct/issues/299

### DIFF
--- a/src/con_duct/suite/plot.py
+++ b/src/con_duct/suite/plot.py
@@ -2,6 +2,7 @@ import argparse
 from datetime import datetime
 import json
 import logging
+import os
 
 lgr = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/con_duct/suite/plot.py
```diff
@@ -23,7 +24,13 @@ def matplotlib_plot(args: argparse.Namespace) -> int:
...
             with open(file_path, "r") as info_file:
                 info_data = json.load(info_file)
-                file_path = info_data["output_paths"]["usage"]
+                rel_usage_path = info_data["output_paths"]["usage"]
+
+                # determine usage abspath
+                rel_info_path = info_data["output_paths"]["info"]
+                abs_info_path = os.path.abspath(file_path)
+                original_wd = abs_info_path[: -len(rel_info_path)].rstrip(os.sep)
+                file_path = os.path.join(original_wd, rel_usage_path)
```
yarikoptic, line 33: managing "paths" manually with `[:]` string operations fragile, and better be avoided. I think the best would be to start using more libpath here... under assumption of them being in the same folder (we have it a "base assumption" so I think it is ok to just add `assert` if really paranoid), could even be as simple as a single `.with_name`.

```
In [24]: p1 = Path('/a/b/c/d/e/f'); rp1 = Path('e/f'); rp2 = Path('e/g')

In [25]: assert rp1.parent == rp2.parent

In [26]: p1.with_name(rp2.name)
Out[26]: PosixPath('/a/b/c/d/e/g')
```

PR #305, opened 2025-09-11
enh: Support duct config files

### LINKED ISSUE #279: Add ability to read from duct config
As the list of CLI flags grows, repeating them across multiple duct commands (or in automation scripts) gets unwieldy. It would be cleaner to load common settings from a dedicated duct config instead of passing every option on the command line.

Read config in the following order, starting with lowest and proceeding to highest precedence: 
- `$XDG_CONFIG_HOME/duct/config.yml`
 - `.duct/config.yml` (project root)
 - `-c path/to/config.yml`
 - CLI flags

asmacdo:
@yarikoptic this is a WIP, but non-draft since this could use design review before continuing into the weeds

- Fixes #279

TODO (adjust) At the high level:
 - startup using `DEFAULT_CONFIG` defined in `__main__`
 - retrieve `-C --config` first if passed
 - load in DEFAULT_CONFIG and then override from each file in order (colon separated)
 - use environment variables if they are present (this is also reflected in the config object
 - the config is then used by the CLI to set default values, which the user can override

NOTE! If you use config files, duct --help will show options' default values to be the ones specified in the config, which I think makes sense but a little unusual.

TODO:
 - [ ] fix tests 
 - [ ] add tests
 - [ ] add to con-duct CLI
 - [ ] make sure -C --config shows up in helptext (currently not)
 - [ ] reconsider behavior "fail if user-provided paths DNE, but pass if default paths DNE" maybe just try and proceed with DEBUG
 - [ ] hand-test thoroughly
 - [ ] optional: use `config` everywhere instead of `args`. IMO nicer usage, but uglier code and a substantially large patch so not worth it
 - [ ] optional: add `con-duct init-conf` to scaffold a config.json file with all default values.

### DIFF
--- a/src/con_duct/__main__.py
+++ b/src/con_duct/__main__.py
@@ -1060,7 +1060,6 @@ class Config:
 
         env_vals, env_src = self._load_env()
         merged, provenance = self._merge_with_provenance(
-            defaults_as_source=True,
             file_layers=file_layers,
             env_vals=env_vals,
             env_src=env_src,
@@ -1136,7 +1135,6 @@ class Config:
 
     def _merge_with_provenance(
         self,
-        defaults_as_source: bool,  # TODO remove this
         file_layers: List[Tuple[Dict[str, Any], str]],
         env_vals: Dict[str, Any],
         env_src: Dict[str, str],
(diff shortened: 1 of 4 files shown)


yarikoptic:
I think we decided to go with jsonargparse and thus this PR could be closed

PR #308, opened 2025-09-13
Add --reverse option to con-duct ls command

yarikoptic:
The -r/--reverse flag lists entries in reverse order, showing the most recent entries on top. This is useful for quickly viewing the latest execution results when there are many entries.

🤖 Generated with [Claude Code](https://claude.ai/code)


asmacdo:
Alternative idea --sort-by, if starts with !, reverse?

yarikoptic:
I stopped liking `-r` since may be at some point we will add `--recursive` to list all from under subfolders (when we get global registry etc)... so may be just `--reverse`.  `--sort-order` might be good to have indeed... let's see if claude would manage...

PR #327, opened 2025-10-20
Combine clis

asmacdo:
This PR combines 2 CLIs into a single CLI for simplicity, maintainability, and (as we learned when adding config) extensibility. 

Prior to this PR the project was separated into 2 CLIs, maintained separately, `duct` and `con-duct`. This worked fine, but IMO was always a little awkward. However problems and complexity have crept up:
 - Duplication: ie logging and arguments for multiple commands 
 - Configuration: When adding config files, (not included in this PR) complexity arose-- sharing configuration files between the CLIs was awkward, and not intuitive.
 
This PR sets us up to:
 - improve separation and organization of the tests: (WIP) https://github.com/con/duct/pull/329
 - setup a consistent  and simple config (WIP) https://github.com/con/duct/pull/333
 
Note to the reviewer: 26 files changed is not scary! Most files are just changing imports.
 - All CLI has been moved to cli.py
 - src/con_duct/__main__.py has been moved to src/con_duct/duct_main.py

Breaking Change:
duct_main.py cannot be directly executed anymore. If users wish to use `duct` without installing `con-duct`, they can still do so by fetching `duct_main.py` and executing their own python using `duct_main.execute()`. 

TODO:
- [x] provide overall summary above on what was Rationale (Problems it would address), How it was done (summary), and what it would open for opportunities
- [x] address copilot's review
- [x] invite others to review
- [ ] rebase other PRs on top before merge so we have it all "ready" and foresee/address problems


src/con_duct/cli.py
```diff
@@ -0,0 +1,474 @@
...
+  DUCT_OUTPUT_PREFIX: see --output-prefix
+  DUCT_SUMMARY_FORMAT: see --summary-format
+  DUCT_SAMPLE_INTERVAL: see --sample-interval
+  DUCT_REPORT_INTERVAL: see --report-interval
+  DUCT_CAPTURE_OUTPUTS: see --capture-outputs
+  DUCT_MESSAGE: see --message
+"""
+
+
+class CustomHelpFormatter(argparse.ArgumentDefaultsHelpFormatter):
```
yarikoptic: please describe here what is "its function" as why we would like to use it.

src/con_duct/cli.py
```diff
@@ -0,0 +1,474 @@
...
+                "--report-interval must be greater than or equal to --sample-interval.",
+            )
+
+    def execute(self) -> int:
+        """Execute the command with these arguments.
+
+        This is a convenience method for tests and library usage.
+        """
+        return duct_execute(
+            command=self.command,
```
yarikoptic: is there 1-to-1 correspondence between `RunArguments` and ArgumentParser's collection of attributes?

ATM it feels very heavy in how much of `X=Y.X` there are -- refactor/get helper to just use a loop with `getattr` here, in `from_argv` and `run_command` .

src/con_duct/cli.py
```diff
@@ -0,0 +1,474 @@
...
+    else:
+        logging.basicConfig(
+            format="%(asctime)s [%(levelname)-8s] %(name)s: %(message)s",
+            datefmt="%Y-%m-%dT%H:%M:%S%z",
+            level=log_level,
+        )
+
+
+@dataclass
+class RunArguments:
```
yarikoptic: so we will have only `Run` arguments but nothing similar for any other command? 

If not, since we are not generally adhering to such design pattern, and now seems to be used only in the tests , where could simply be used via `args` directly. That would also minimize amount of all those heavy manual listing of all args in each different kind of invocation

src/con_duct/cli.py
asmacdo: after removing RunArguments this breaks down to just removing extra args and passing kwargs

CLAUDE.md
```diff
@@ -3,8 +3,7 @@
 This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
 
 ## Build & Test Commands
-- Use `python -m venv .venv` to create a virtual environment
-- Use `source .venv/bin/activate` to activate a virtual environment
+- A virtual environment should be used (see .local.CLAUDE.md for venv location)
```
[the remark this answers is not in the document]
yarikoptic: should be git ignored and instruct "if present" I guess so my claude would not get confused?
 You can also generally configure such things in your `~/.claude/CLAUDE.md` , or at least this is what I did:

```
❯ head ~/.claude/CLAUDE.md
# Some notes on Python projects

- I used to use `venv` with development libraries installed under
  venvs/dev{python_version}, e.g. venvs/dev3.  Now I tend to use `uv`, so
  I use `uv venv` and `uv pip install` and `uv run`.   If needed can
  `source .venv/bin/activate` to enter/use such venv.

- I also tend to like `tox` which I use with `tox-uv` to speed things up.
  So for new projects, set them up with tox.ini and in existing use `tox -e py3` etc.
```

setup.cfg
```diff
@@ -66,8 +66,8 @@ all =
 
 [options.entry_points]
 console_scripts =
-    duct = con_duct.__main__:main
-    con-duct = con_duct.suite.main:main
+    duct = con_duct.cli:duct_entrypoint
+    con-duct = con_duct.cli:main
```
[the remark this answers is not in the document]
asmacdo: no plan to deprecate, `duct` is now just be a shortcut for `con-duct run`, and they should behave exactly the same.

CodyCBakerPhD:
> Note to the reviewer: 26 files changed is not scary! Most files are just changing imports.

`#` of files changed, not so scary

+900/-800 lines changed always scary 😆what if 1 important line missing

@asmacdo Aside from trusting tests, how best do you recommend testing this PR works? Ensuring the rebased config/other PRs work well? Trying a couple commands locally?

Also what is your opinion on these codecov CI checks

setup.cfg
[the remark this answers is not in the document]
CodyCBakerPhD: OK - towards the eternal spirit of [PEP 20 (Zen of Python - line 12 - "There should be one-- and preferably only one --obvious way to do it.")](https://peps.python.org/pep-0020/) perhaps put some thought into going one way or another to remove confusion, reduce number of things to maintain, and remove a source of possible bugs (if wires get crossed or imports/invocations get changed)

CodyCBakerPhD:
> Coverage seems fine IMO-- we don't have tests for CLI parsing, since thats a big chunk of what got moved, the patch% is pretty low. I dont love that its not covered, but testing argparse usage doesnt add much value IMO (interested to hear if you disagree)

I meant do you find it useful that they show as failing on the PR checks

CLAUDE.md
```diff
@@ -3,8 +3,7 @@
 This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
 
 ## Build & Test Commands
-- Use `python -m venv .venv` to create a virtual environment
-- Use `source .venv/bin/activate` to activate a virtual environment
+- A virtual environment should be used (see .local.CLAUDE.md or ~/.claude/CLAUDE.md if they exist for venv locations)
```
yarikoptic: Below you use @ for referencing files but not here

yarikoptic (approved):
Left one non consequential comment, let's see it on action

PR #333, opened 2025-10-30
Add dotenv config

asmacdo:
# Add .env File Support for Configuration

Fixes: 
 - https://github.com/con/duct/issues/279

## Summary

Adds support for loading duct configuration from `.env` files, making it easier to manage default settings across projects and environments without cluttering shell profiles or requiring command-line flags.

## Key Features

- **Multiple config paths**: uses 3 default locations, overridable with DUCT_CONFIG_PATHS. Defaults:
    -   /etc/duct/.env                              # System-wide defaults
    -   ${XDG_CONFIG_HOME:-~/.config}/duct/.env     # User defaults
    -  .duct/.env                                  # Project-specific
- **Proper precedence**: CLI args > explicit env vars > .env files > defaults
- **Graceful degradation**: Works without python-dotenv (skips .env loading)
- **Robust error handling**: Malformed, unreadable, or missing files are logged and skipped
- **Variable expansion**: Supports `${VAR:-default}` syntax in paths
- **Deferred logging**: .env loading happens before logger setup, messages are buffered and replayed
- Loaded .env files are logged at INFO level for visibility

## TODO Before Merge

- [x] Wait for #327 (combine-cli) to be merged
- [x] Rebase onto main after #327 merge
- [x] Mark as ready for review

**Note:**  .env values overrides the defaults, and will show up as the default in helptext

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,6 +45,9 @@ jobs:
           - python-version: '3.10'
             toxenv: typing
             os: ubuntu-latest
+          - python-version: '3.10'
+            toxenv: minimal
+            os: ubuntu-latest
     steps:
       - name: Check out repository
         uses: actions/checkout@v4


yarikoptic:
dotenv is an optional dependency... Do we have a test run verifying correct operation of basic duct without it?

Issue #342, opened 2025-12-06
Use .jsonl (not .json) extension for json lines file(s) such as `_usage.<ext>`


yarikoptic:
ATM we use regular `.json` extension which is misleading since it is not JSON but JSON lines (https://jsonlines.org).

That complicates e.g. creating proper (not "_usage"-suffix dependent) solutions like in

- https://github.com/con/duct/pull/341

We need to refactor to

- use `.jsonl` whenever we use json lines file
- keep code backward compatible to be able to understand/load already existing records which have regular `.json` extension
- code #341 to also support any `.jsonl` file as "json lines"
- ideally centralize the logic of `load_jsonbased` (or alike, looking at #341 we do not have that centralized) which would tell apart two kinds of files based on extension (or `_usage` suffix ATM as well)

Issue #339, opened 2025-12-05
Bug with pretty print


CodyCBakerPhD:
@asmacdo Can you replicate?

```bash
mkdir test1
cd test1

duct sleep 5
con-duct pp .duct/logs/<path>_usage.json
> con_duct.pprint_json: Error decoding JSON: Extra data: line 2 column 1 (char 418)
```

I assume this isn't something introduced in #327 since I did not see any core changes to that logic altered

asmacdo:
<details><summary>replication</summary>

```
 venv  /tmp/duct   combine-clis  duct sleep 5
2025-12-05T12:36:34-0600 [INFO    ] con-duct: duct 0.17.0.post56+g5246d6d is executing 'sleep 5'...
2025-12-05T12:36:34-0600 [INFO    ] con-duct: Log files will be written to .duct/logs/2025.12.05T12.36.34-1257662_
2025-12-05T12:36:39-0600 [INFO    ] con-duct: Summary:
Exit Code: 0
Command: sleep 5
Log files location: .duct/logs/2025.12.05T12.36.34-1257662_
Wall Clock Time: 5.010 sec
Memory Peak Usage (RSS): 2.4 MB
Memory Average Usage (RSS): 2.4 MB
Virtual Memory Peak Usage (VSZ): 235.9 MB
Virtual Memory Average Usage (VSZ): 235.9 MB
Memory Peak Percentage: 0.00%
Memory Average Percentage: 0.00%
CPU Peak Usage: 0.00%
Average CPU Usage: 0.00%

 venv  /tmp/duct   combine-clis  con-duct pp 
usage: con-duct pp [-h] [-l {NONE,CRITICAL,ERROR,WARNING,INFO,DEBUG}] [-q] [-H] file_path
con-duct pp: error: the following arguments are required: file_path
 ✘  venv  /tmp/duct   combine-clis  con-duct pp .duct/logs/2025.12.05T12.36.34-1257662_info.json 
{'command': 'sleep 5',
 'duct_version': '0.17.0.post56+g5246d6d',
 'env': {},
 'execution_summary': {'average_pcpu': 0.0,
                       'average_pmem': 0.0,
                       'average_rss': 2387968.0,
                       'average_vsz': 235872256.0,
                       'command': 'sleep 5',
                       'end_time': 1764959799.219335,
                       'exit_code': 0,
                       'logs_prefix': '.duct/logs/2025.12.05T12.36.34-1257662_',
                       'num_reports': 2,
[...]

PR #344, opened 2025-12-08
Use jsonl suffix and centralize json loading

asmacdo:
Solves 2 issues: 
 - Fixes https://github.com/con/duct/issues/342
 - Fixes https://github.com/con/duct/issues/339
 
 Changes:
- Change usage file extension from .json to .jsonl for new runs
- Add usage_legacy suffix for backward compatibility with existing files
- Create json_utils.py with centralized file type detection and loading
- Update pprint, plot, ls to use json_utils functions
- Update tests to use SUFFIXES constants instead of hardcoded strings


asmacdo:
closing-- trying a PR from con fork to see if I can get copilot review https://github.com/con/duct/pull/345

PR #351, opened 2025-12-20
Added Mac (M-series) support

CodyCBakerPhD:
Redo of #325 under the new design with minimal adjustments for Mac support

The tests as written run fine and reliably on my local device, but we'll see what the CI says

I did add a specific behavior for skipping samples during aggregation to pass a couple of tests, but I emphasize that the behavior of the tool on my system (when I run the same sleep command through terminal) appears to be working as intended, and the reason the samples fail to get captured is unknown (which appear to be the final samples captured in the process)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         os:
           # - macos-12
-          - macos-latest
+          - macos-latest  # arm64
           # - windows-latest
           - ubuntu-latest
         python-version:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,8 +9,6 @@ build-backend = "setuptools.build_meta"
 [tool.versioningit.write]
 file = "src/con_duct/_version.py"
 
-
-
 [tool.pytest.ini_options]
 markers = [
     "flaky: mark a test as being unreliable"


src/con_duct/cli.py
```diff
@@ -366,6 +367,15 @@ def _create_run_parser() -> argparse.ArgumentParser:
         "'current-session' tracks the current session instead of starting a new one. "
         "Useful for tracking slurm jobs or other commands that should run in the current session.",
     )
+    parser.add_argument(
+        "--skipempty",
+        action="store_true",
+        help=(
+            "Skip sample aggregation if samples fail to be collected. "
+            "Default is True for macOS and False for Linux."
```
candleindark:
```suggestion
            "Default is True for macOS and False for otherwise."
```

candleindark (changes requested):
Some minor suggestions at this point.

src/con_duct/duct_main.py
```diff
@@ -20,11 +21,22 @@
...
 from types import FrameType
-from typing import IO, Any, Optional, TextIO
+from typing import IO, Any, Callable, Optional, TextIO, Union
+import warnings
 
 __version__ = version("con-duct")
 __schema_version__ = "0.2.2"
 
+is_mac_intel = sys.platform == "darwin" and os.uname().machine == "x86_64"
+if is_mac_intel and not os.getenv("DUCT_IGNORE_INTEL_WARNING"):
```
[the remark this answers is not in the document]
CodyCBakerPhD: Yeah, probably need something like https://github.com/symonsoft/str2bool/blob/master/str2bool/__init__.py#L3-L18 for that

src/con_duct/cli.py
CodyCBakerPhD: There is no Windows support yet so I was thinking to be explicit until then

CodyCBakerPhD: It might not be a problem for Windows when I add that, thus it might not be true in the 'otherwise' case at that time, in which then *I* would have to remember that inconsistency ;)

candleindark: I get the sense that we may not be talking about the same thing.

My concern about the help message is that it is inconsistent with how the default for this option is defined. The default for this option is set to `SKIPEMPTY_DEFAULT` which is set to `False` when the system not a Mac as in https://github.com/CodyCBakerPhD/duct/blob/da604be8cf0fde2a59c482b413628255cb1652ce/src/con_duct/duct_main.py#L39. Regardless if Windows has failures in collecting samples, `SKIPEMPTY_DEFAULT` and the default of this option is `False` in Windows.

Taking the suggested change ensure the help message and the definition of the default of this option are consistent. If one needs to change the definition the default of this option, or `SKIPEMPTY_DEFAULT`,  in the future, one would naturally change the help message.

candleindark: If not taking the suggestion is not to suggest an unsupported platform and its possible behavior. I agree with the choice.

src/con_duct/duct_main.py
```diff
@@ -321,6 +341,128 @@ def for_json(self) -> dict[str, Any]:
...
+
+    lines = _get_ps_lines_mac()
+    pid_to_matching_sid = {
+        pid: sid
+        for line in lines
+        if (sid := _try_to_get_sid(pid=(pid := int(line.split(maxsplit=1)[0]))))
+        == session_id
+    }
+
+    if not pid_to_matching_sid:
```
asmacdo: Are we missing samples during regular execution or does this only occur at the end when the top process has finished?

If we don't have any samples, we should return `None`, which indicates that the process has probably finished (unless we have actually missed a sample, but AFAIK thats only hypothetical). What happens in this code is that an empty sample is returned instead, which is why the skipempty stuff is needed later. I think if we just returned None here like the linux sample collection, that would allow duct to exit faster after everything is finished (or just skip the missed sample altogether). https://github.com/con/duct/blob/da604be8cf0fde2a59c482b413628255cb1652ce/src/con_duct/duct_main.py#L832-L842 Note: we should log that a sample was missed before process has ended though. 

With that change I think we can drop the skipempty stuff.

src/con_duct/duct_main.py
```diff
@@ -321,6 +341,128 @@ def for_json(self) -> dict[str, Any]:
...
+        "-o",
+        "pid,pcpu,pmem,rss,vsz,etime,stat,args",
+    ]
+    output = subprocess.check_output(ps_command, text=True)
+
+    lines = [line for line in output.splitlines()[1:] if line]
+    return lines
+
+
+def _add_sample_from_line_mac(
```
asmacdo: could we rename to `_add_pid_to_sample_from_line_mac` or similar? This isnt actually adding a `sample`, its adding a partial sample.

asmacdo (changes requested):

src/con_duct/duct_main.py
CodyCBakerPhD: Done

src/con_duct/duct_main.py
```diff
@@ -321,6 +341,128 @@ def for_json(self) -> dict[str, Any]:
...
+                line=line, pid_to_matching_sid=pid_to_matching_sid, sample=sample
+            )
+            for line in lines
+        ),
+        maxlen=0,
+    )
+
+    try:
+        sample.averages = Averages.from_sample(sample=sample)
+    except AssertionError:
```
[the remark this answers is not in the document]
CodyCBakerPhD: Depends on your architectural determination for how to handle the missing samples case

As it stands, this is where the main blocking error occurs in that event ATM

src/con_duct/duct_main.py
asmacdo: Done. https://github.com/CodyCBakerPhD/duct/pull/1 That PR is a bit blind since I dont have a mac, but the tests pretty much passed except for 1 previously known flake. (For the flake, I'll come back to that once we get the architecture ironed out).

CodyCBakerPhD: Thanks!

src/con_duct/duct_main.py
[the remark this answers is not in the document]
CodyCBakerPhD: Included in infrastructure PR

src/con_duct/duct_main.py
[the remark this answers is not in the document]
CodyCBakerPhD: Done

src/con_duct/duct_main.py
```diff
@@ -321,6 +332,123 @@ def for_json(self) -> dict[str, Any]:
...
+        lgr.debug(f"Error fetching session ID for PID {pid}: {str(exc)}")
+        return -1
+
+
+def _get_ps_lines_mac() -> list[str]:
+    ps_command = [
+        "ps",
+        "-ax",
+        "-o",
+        "pid,pcpu,pmem,rss,vsz,etime,stat,args",
```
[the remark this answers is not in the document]
CodyCBakerPhD: Unfortunately, that is precisely the core issue with the mac implementation of `ps` (no built-in `sid` filtering), and awareness of this goes back to the original issue for this support

The current approach is the ONLY currently known way to approximate the behavior. It is not as efficient, clearly

CodyCBakerPhD:
CI runs look to be pretty stable now as well!

src/con_duct/duct_main.py
```diff
@@ -321,6 +351,123 @@ def for_json(self) -> dict[str, Any]:
...
+    return sample
+
+
+def _try_to_get_sid(pid: int) -> int:
+    """
+    It is possible that the `pid` returned by the top `ps` call no longer exists at time of `getsid` request.
+    """
+    try:
+        return os.getsid(pid)
+    except Exception as exc:
```
asmacdo: I think we can tighten this up a bit
```suggestion
    except OSError as exc:
```

asmacdo (commented):
I'll have another look tomorrow morning with fresh eyes. Just a couple nitpicks at this point, I think this is almost ready.

Thanks for all this effort @CodyCBakerPhD <3

src/con_duct/duct_main.py
CodyCBakerPhD:
```suggestion
    except ProcessLookupError as exc:
```

CodyCBakerPhD: (which I have now adjusted)

.github/workflows/test.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
       matrix:
         os:
           # - macos-12
-          # - macos-latest
+          - macos-latest
```
yarikoptic: this is an arm one, and I [see some warning](https://github.com/con/duct/pull/351/files#diff-faaff2e5a844ecc33a43feb37cf7d0ea416c8c9b2846040c742f7534d7b8d9ceR49) to be issued on intel below. Should we may be add

```suggestion
          - macos-latest        # arm64
          - macos-latest-large  # intel
```

to test on both?

CodyCBakerPhD: I don't believe our organization use large runners unless you upgraded the plan

`macos-15-intel` would be our next bet but I would just test in a follow-up since this has gone on long enough

Issue #354, opened 2026-01-05
Elaborate filterwarnings instead of capping versions of depends


yarikoptic:
Follow up to

- https://github.com/con/duct/pull/353/files

which precludes us detecting new possible gotcha's... IMHO we better just filter away that warning for now -- see below we have just `filterwarnings = error` -- thus error, and we just to ignore that particular one. See how done in https://github.com/dandi/dandi-cli/blob/master/tox.ini#L47

PR #355, opened 2026-01-05
filter pyparsing deprecation warnings instead of pinning

asmacdo:
Fixes https://github.com/con/duct/issues/354

Replace the pyparsing version pin with a filterwarnings ignore for matplotlib's deprecated pyparsing API usage. This approach allows detecting new issues from dependency updates while suppressing known-acceptable warnings.

Supersedes the approach in #353.


yarikoptic:
green is good!

Issue #313, opened 2025-09-26
Add a mode to "interleave" stdout and stderr


yarikoptic:
ATM we write both stdout and stderr into different files.

I would like to be able to switch to another mode, which would dump them into the same file, so ideally mimicing how humans see it dumped to the terminal. Ideally we might want to annotate somehow which stream it is e.g.

```
[stdout] output to stdout
[stderr] output to stderr
```

may be there is already a separate helper which could help us with that and push everything to stdout with annotations (not just `2>&1`) and we could just turn off e.g. stderr output, and overall just an example/howto?

asmacdo:
**Interface options:**

**A) Extend existing enum:**
```
-c {all,none,stdout,stderr,combined,annotated}
```

**B) New format flag:**
```
--capture-format {separate,combined,annotated}
```

**C) Boolean flags:**
```
--interleave              # merge into single file
--annotate-streams        # add [stdout]/[stderr] prefixes (implies --interleave)
```

Output file: `{prefix}output` (instead of separate `stdout`/`stderr`)

asmacdo:
https://github.com/con/duct/issues/213 for timestamp prepend

candleindark:
As a note, there is no interleaving of stdout and stderr outputs even in the printout to the terminal at this point.

<details>
<summary>interleave.py</summary>

```py
import sys
import time


def main() -> None:
    # Hardcoded number of interleaving iterations.
    n = 20

    for i in range(n):
        # Write to stdout then stderr, flushing each time to make interleaving visible.
        sys.stdout.write(f"stdout {i}\n")
        # sys.stdout.flush()

        # Tiny delay to make alternating output easier to observe in terminals/process capture.
        # time.sleep(0.01)

        sys.stderr.write(f"stderr {i}\n")
        # sys.stderr.flush()

        # time.sleep(0.01)


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary>duct python interleave.py</summary>

```console
2026-01-16T10:26:01-0800 [INFO    ] con-duct: No .env files found
2026-01-16T10:26:01-0800 [INFO    ] con-duct: duct 0.18.0.post11+gda604be is executing 'python interleave.py'...
2026-01-16T10:26:01-0800 [INFO    ] con-duct: Log files will be written to .duct/logs/2026.01.16T10.26.01-24632_
stdout 0
stdout 1
stdout 2
stdout 3
stdout 4
stdout 5
stdout 6
stdout 7
stdout 8
stdout 9
[...]

PR #370, opened 2026-01-16
WIP: Investigate PTY-based TeeStream for interleaving (#313)

asmacdo:
## Summary

This PR reintroduces the PTY-based `TeeStream` class (which was replaced by `TailPipe` in May 2024) to investigate whether it can help with stdout/stderr interleaving for #313.

### The buffering problem

When capturing output through duct, interleaving is lost even when the child process outputs in order:

```
# Expected (what child writes):
stdout 0, stderr 0, stdout 1, stderr 1, ...

# Actual (what duct shows):
stdout 0-19, then stderr 0-19
```

### What we tested

| Approach | Without duct | Through duct |
|----------|-------------|--------------|
| Base case | ❌ batched | ❌ batched |
| PYTHONUNBUFFERED=1 | ✅ interleaved | ❌ still batched |
| python -u | ✅ interleaved | ❌ still batched |
| stdbuf -oL | ✅ interleaved | ❌ still batched |

**Key insight:** The child's buffering isn't the problem - duct's architecture is.

### Why TailPipe doesn't interleave

1. `Popen(stdout=file1, stderr=file2)` - streams separated at kernel level
2. Two independent TailPipe threads read from separate files
3. Thread scheduling determines display order, not write order

### Why TeeStream (PTY) also doesn't interleave (in current form)

We reintroduced TeeStream with PTY-based capture, but with **two separate PTYs** (one per stream), interleaving is still lost.

### What DOES work

When both stdout AND stderr go to the **same PTY slave fd**:

```python
master_fd, slave_fd = os.openpty()
proc = subprocess.Popen(
    ["python", "interleave.py"],
    stdout=slave_fd,
    stderr=slave_fd,  # Both to same PTY!
)
# Read from master_fd -> perfect interleaving!
```

This gives perfect interleaving because the kernel merges both streams in arrival order.

## Recommendation for #313

For a `--capture-outputs=combined` or `annotated` mode:
- Use a **single shared PTY** for both stdout and stderr
- Read from the PTY master and write to a combined output file
- Optionally annotate with `[stdout]`/`[stderr]` prefixes (though with shared PTY, we lose stream distinction)

If stream distinction is needed with interleaving, would need the pipe+select approach:
- Use `subprocess.PIPE` for both streams
- Use `select.select()` to multiplex reads in a single thread
- Annotate each chunk with its source stream

## Test script

`interleave.py` is included for reproducing the behavior.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/con_duct/duct_main.py
+++ b/src/con_duct/duct_main.py
@@ -882,40 +882,55 @@ class TailPipe:
 class TeeStream:
     """TeeStream uses a PTY to capture output with proper line-buffering.
 
+    PTY data flow:
+        subprocess writes to pty_subprocess_fd (appears as a terminal to child)
+                            ↓
+                    [kernel PTY layer]
+                            ↓
+        duct reads from pty_read_fd → writes to both file and terminal buffer
+
     Unlike TailPipe which reads from a file, TeeStream uses a pseudo-terminal
-    which causes the child process to use line-buffered output by default.
-    This preserves interleaving order when both stdout and stderr are captured.
+    which causes the child process to use line-buffered output by default
+    (because it thinks it's writing to a terminal).
+
+    Note: For true stdout/stderr interleaving, both streams must share the
+    same PTY. With separate TeeStream instances per stream, interleaving
+    is still lost. See #313 for discussion.
     """
 
     def __init__(self, file_path: str, buffer: IO[bytes]) -> None:
         self.file_path = file_path
         self.buffer = buffer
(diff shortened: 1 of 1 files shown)


asmacdo:
@candleindark this is quickly-generated AI outputs, so its not fully trustworthy, but it aligns with my understanding & memory of how things once were. Hope this is helpful, I'm down to pair on this too if you like.

Issue #304, opened 2025-09-11
deprecate {datetime_filesafe} and prefer {datetime} for output prefix


asmacdo:
Its overly verbose, and since its part of a filename, filename-safety can be assumed.

PR #375, opened 2026-01-23
feat: replace support of "datetime_filesafe" format field in output file prefix format string with "datetime"

candleindark:
This PR closes https://github.com/con/duct/issues/304

### DIFF
--- a/test/duct_main/test_log_paths.py
+++ b/test/duct_main/test_log_paths.py
@@ -14,6 +14,14 @@ def test_log_paths_datetime_prefix() -> None:
         assert re.match(pattern, path) is not None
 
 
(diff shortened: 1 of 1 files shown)


asmacdo:
yep we'll need to rebase (i moved this). either I will do or @candleindark on Friday?

candleindark:
@asmacdo Have you seen the error at https://github.com/con/duct/actions/runs/21525492435/job/62027909892?pr=375 before? It seems to be an unrelated issue.

Screenshot 2026-01-30 at 10 18 39 AM


It can be an intermittent failure, but I can't rerun tests in this repo to find out if that's the case.

asmacdo:
@candleindark yep I think this one just needs a rerunflake back off https://github.com/con/duct/issues/384

PR #403, opened 2026-04-04
WIP: Validate resource stats match real usage

asmacdo:
## [Update 4/21]

Claude-code agents have discovered 2-3 correctness issues related to how we do sampling and aggregation, but its complex enough I am going to need a dedicated deep focus block.  Rather than [slopsplaining](https://www.urbandictionary.com/define.php?term=slopsplaining) here, instead I'll update after I dig in a little more.

tldr; These tests might be useful, but they aren't sufficient to catch the underlying issues surfaced in the original issue.

Related
   - https://github.com/con/duct/issues/399

## Summary

- Add integration tests that run programs with **known, predictable resource usage** and verify duct's measurements match reality (not just aggregation math with synthetic data)
- Single-process tests: memory allocation detected in peak_rss, wall clock accuracy, idle vs busy CPU
- **Child/forked process tests**: spawn N children each allocating M MB, verify total RSS reflects the sum, per-process stats track individual children, and totals equal sum of per-process values
- Add `test/data/memory_children.py` helper using `multiprocessing.Process`

## Test plan

- [x] All 9 new tests pass locally (`test/duct_main/test_resource_validation.py`)
- [x ] CI passes
- [x] @asmacdo reviews for suitability and flakiness tolerance

https://claude.ai/code/session_01FYRR4Y8PzNLBU344ovdDS5

### DIFF
--- a/test/data/memory_children.py
+++ b/test/data/memory_children.py
@@ -6,8 +6,8 @@ Usage: memory_children.py <num_children> <mb_per_child> <hold_seconds>
 Each child allocates mb_per_child MB and holds it for hold_seconds.
 The parent waits for all children to finish.
 """
-from __future__ import annotations
 
+from __future__ import annotations
 import multiprocessing
 import sys
 import time
--- a/test/data/memory_children.py
+++ b/test/data/memory_children.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+"""Spawn child processes that each allocate a known amount of memory.
+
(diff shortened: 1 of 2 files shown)


yarikoptic:
I thought we had an issue about the odd resource stats reported, but I am failing to find it and do not see a ref here. do you remember it @asmacdo ?

asmacdo:
Some of these tests are useful but not all. Rolling this up into a future PR.

asmacdo:
in the description update but for conversation clarity, the odd resource stats were reported in #399

PR #413, opened 2026-04-20
pp: humanize memory_total under -H

### LINKED ISSUE #412: con-duct pp -H does not humanize memory_total
## Summary

`con-duct pp -H` humanizes most byte-count fields but misses `memory_total`. Users expect every byte-count under `-H` to be humanized.

## Reproducer

```sh
duct -p /tmp/p_ --sample-interval 0.2 --report-interval 0.5 -- sleep 2
con-duct pp -H /tmp/p_info.json | grep -E "memory_total|peak_rss"
```

## Observed

```
'peak_rss': '10.7 MB',           # humanized
'memory_total': 33330020352,     # raw bytes
```

## Expected

`memory_total` humanized like `peak_rss` / `peak_vsz`:

```
'memory_total': '31.0 GB',
```

## Fix direction

`src/con_duct/pprint_json.py`: the humanize path applies per-key heuristics; `memory_total` isn't in the list. Add it.

## Scope

Reproduces on v0.19.0 and current `main`. Longstanding, not a regression. Surfaced during pre-0.20.0 QE.

asmacdo:
`con-duct pp -H` humanizes byte-count fields like `peak_rss` and
`peak_vsz` but missed `memory_total` (under `system`), so users saw a
raw integer like `1081801371648` next to humanized siblings.

Adds `memory_total` to the per-key conversion mapping in
`pprint_json.get_field_conversion_mapping()` (uses the `S` size
conversion) and extends `test_apply_conversion_with_numbers` to cover
it.

Closes con/duct#412.

### DIFF
--- a/src/con_duct/pprint_json.py
+++ b/src/con_duct/pprint_json.py
@@ -20,6 +20,7 @@ def get_field_conversion_mapping() -> dict[str, str]:
         "average_vsz": "S",
         "peak_pcpu": "P",
         "peak_pmem": "P",
+        "memory_total": "S",
         "peak_rss": "S",
         "peak_vsz": "S",
         "wall_clock_time": "T",
--- a/test/test_pprint.py
+++ b/test/test_pprint.py
@@ -84,6 +84,11 @@ class TestPPrintHumanization(unittest.TestCase):
         assert result == "1.0 MB"
 
         result = pprint_json._apply_conversion(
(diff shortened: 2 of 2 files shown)


asmacdo:
@CodyCBakerPhD picked this one since its tiny and its been annoying me. My primary intent is just to cut a duct release.

yarikoptic (approved):

PR #414, opened 2026-04-23
Document what duct's resource statistics actually measure

asmacdo:
Explain the semantics of each field (`pcpu`, `rss`, `pmem`, `vsz`) under duct's current `ps`-based sampler, where the numbers are reliable, and where they mislead. Motivated by recurring confusion: `peak_pcpu` exceeding `Ncores × 100%` (issue #399), `peak_vsz` reaching terabytes on workloads with only GBs of real memory, and `total_*` disagreeing with `sum(per-pid)` in aggregated records. All are correct behavior of `ps` semantics and duct's max-reduction aggregation — not bugs — but there was no reference explaining why.

Scope is current behavior only; alternative samplers (psutil, cgroup) will be documented separately.

TODO:
 - [ ] consider alternatives, start design doc PR and potentially reference here for discussion
 - [ ] 1 line per sentence, other cleanup/human review
 - [ ] Create some scenarios and/or tests to demonstrate this behavior

### DIFF
--- a/docs/resource-statistics.md
+++ b/docs/resource-statistics.md
@@ -1,0 +1,483 @@
+# Interpreting duct's resource statistics
+
+duct records resource usage in two places:
+
+- **`usage.jsonl`** — one JSON record per report interval (default:
+  every 60 seconds), capturing per-process and session-total stats
+  aggregated over that window.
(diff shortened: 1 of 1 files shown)


asmacdo:
Superseded by https://github.com/con/duct/pull/424

PR #420, opened 2026-04-25
fix: clean error on output prefix collision instead of traceback

### LINKED ISSUE #410: Clobber collision surfaces raw Python traceback instead of clean error
## Summary

Running `duct` against an output prefix that already has files (no `--clobber`) surfaces a full Python traceback, even though the underlying `FileExistsError` carries a perfectly good user-facing message.

## Reproducer

```sh
duct -p /tmp/c_ -- echo first     # succeeds
duct -p /tmp/c_ -- echo second    # fails with traceback
```

## Observed

```
Traceback (most recent call last):
  ...
FileExistsError: Conflicting files:
- /tmp/c_stdout
- /tmp/c_stderr
- /tmp/c_usage.jsonl
- /tmp/c_info.json
Use --clobber to overwrite conflicting files.
```

Exit code: 1.

## Expected

Log the error at ERROR level (the existing message is already good), exit nonzero, no traceback. `FileExistsError` here is an expected user-facing condition, not a crash.

## Fix direction

`src/con_duct/_models.py:152` (`prepare_paths`) raises. Caller is `src/con_duct/_duct_main.py:63` (`execute`). Wrap the `log_paths.prepare_paths(...)` call in a try/except, emit `lgr.error(str(e))`, return non-zero.

## Scope

Reproduces on v0.19.0 and current `main`. Longstanding, not a regression. Surfaced during pre-0.20.0 QE.

lobennett:
Running `duct` (or `con-duct run`) against an output prefix where files already exist — without `--clobber` — prints a full Python traceback ending in `FileExistsError`, exiting 1. The exception's message is already user-facing quality (it lists the conflicting paths and tells the user to pass `--clobber`), but the traceback wrapper makes it look like duct crashed.

Catches `FileExistsError` at the call site of `LogPaths.prepare_paths()` in `_duct_main.execute()`, logs the existing message at ERROR level via `lgr`, and returns 1. Mirrors the existing `FileNotFoundError` handler a few lines below in the same function. The exception and its message in `_models.py` are unchanged; exit code is unchanged (1). Adds `test_execute_existing_files_no_clobber` to `test/duct_main/test_execution.py` mirroring the shape of `test_execute_unknown_command`: covers exit code, ERROR log content, and that the never-created sibling outputs are not spuriously created.

Closes con/duct#410.

### DIFF
--- a/src/con_duct/_duct_main.py
+++ b/src/con_duct/_duct_main.py
@@ -60,7 +60,11 @@ def execute(
         )
 
     log_paths = LogPaths.create(output_prefix, pid=os.getpid())
-    log_paths.prepare_paths(clobber, capture_outputs)
+    try:
+        log_paths.prepare_paths(clobber, capture_outputs)
+    except FileExistsError as e:
+        lgr.error("%s", e)
+        return 1
     stdout, stderr = prepare_outputs(capture_outputs, outputs, log_paths)
     stdout_file: TextIO | IO[bytes] | int | None
     if isinstance(stdout, TailPipe):
--- a/test/duct_main/test_execution.py
+++ b/test/duct_main/test_execution.py
@@ -363,3 +363,17 @@ def test_no_message_in_json_output(temp_output_dir: str) -> None:
 
     assert "message" in info_dict
     assert info_dict["message"] == ""
+
+
+def test_execute_existing_files_no_clobber(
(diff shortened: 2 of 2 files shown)


asmacdo:
@lobennett Thanks for these! I'll try to review today or tomorrow

PR #421, opened 2026-04-25
fix: emit deprecation warning for {datetime_filesafe} format field

### LINKED ISSUE #411: {datetime_filesafe} backward-compat emits no deprecation warning
## Summary

`{datetime_filesafe}` was renamed to `{datetime}` in the January 2026 refactor. The old name is kept as a backward-compat shim with a comment noting "It should be removed eventually" — but users still using `{datetime_filesafe}` get no warning. They won't migrate until the shim is removed, at which point the breakage looks like a regression.

## Reproducer

```sh
duct -p '/tmp/old_{datetime_filesafe}_' -- echo hi
# produces /tmp/old_2026.04.20T15.52.07_* with no warning of any kind
```

## Expected

One of:
- `warnings.warn("{datetime_filesafe} is deprecated, use {datetime}", DeprecationWarning)` on first encounter.
- `lgr.warning(...)` log message.

Either is fine; the point is signalling to users who are scripting against the old name so they can migrate before the shim is removed.

## Fix direction

In `LogPaths.create` (or wherever the output prefix is substituted), inspect `output_prefix` for `{datetime_filesafe}` and emit a warning once per invocation.

## Scope

- current HEAD: `2162da9` (shim added commit `3cdec99`, Jan 2026)
- v0.19.0: n/a (only `{datetime_filesafe}` existed)

Not a regression — it's a deprecation that was added silently. Surfaced during pre-0.20.0 QE. Worth fixing before 0.20.0 so the shim's eventual removal doesn't surprise users.

lobennett:
PR #375 (January 2026) renamed the `{datetime_filesafe}` placeholder in `--output-prefix` to `{datetime}`. The legacy name was retained as a backward-compat shim in `LogPaths.create()` (`_models.py`) with a comment noting "It should be removed eventually" — but users on the legacy name receive no signal that they're on a deprecated placeholder, so they won't migrate before the shim is dropped.

Adds a substring check at the top of `LogPaths.create()`: if `{datetime_filesafe}` appears in `output_prefix`, emit `lgr.warning("{datetime_filesafe} in --output-prefix is deprecated; use {datetime} instead")` once per invocation. Behavior is otherwise unchanged. The existing `test_log_paths_deprecated_datetime_filesafe_prefix` is extended (per datalad CONTRIBUTING.md's "minimal assertion" guidance) to assert the warning text appears in `caplog`.

Closes con/duct#411.

### DIFF
--- a/src/con_duct/_models.py
+++ b/src/con_duct/_models.py
@@ -9,6 +9,7 @@ from enum import Enum
 import logging
 import os
 import re
+import string
 from typing import Any, Optional
 from con_duct._constants import SUFFIXES
 from con_duct._utils import assert_num
@@ -129,6 +130,13 @@ class LogPaths:
 
     @classmethod
     def create(cls, output_prefix: str, pid: None | int = None) -> LogPaths:
+        if any(
+            field_name == "datetime_filesafe"
+            for _, field_name, _, _ in string.Formatter().parse(output_prefix)
+        ):
+            lgr.warning(
+                "{datetime_filesafe} in --output-prefix is deprecated; use {datetime} instead"
+            )
         datetime_filesafe = datetime.now().strftime("%Y.%m.%dT%H.%M.%S")
         formatted_prefix = output_prefix.format(
             pid=pid,
--- a/test/duct_main/test_log_paths.py
+++ b/test/duct_main/test_log_paths.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from dataclasses import asdict
+import logging
 import os
 import re
 from unittest.mock import MagicMock, call, patch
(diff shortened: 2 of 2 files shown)


lobennett:
The `codecov/project` −0.16% is from `_tracker.py:235,242` — inside the `if sample is None:` block in `monitor_process` that handles failed sample reads. That code is unchanged in this PR (`git diff main..HEAD -- src/con_duct/_tracker.py` is empty), and that block only runs when a sample read happens during the subprocess's exit, so its coverage varies between CI runs. Codecov's per-line API confirms: base hit both lines, head shows them as miss/partial. Patch coverage is 100%.

asmacdo:
Thanks @lobennett looks good to me.

PR #424, opened 2026-04-29
Add plot --cpu modes, including ps-cpu-timepoint for derived instantaneous CPU

asmacdo:
## Summary

What this PR does is rework the plot so the data is presented honestly.
The old chart drew `totals.pcpu` and `totals.rss` as single lines with no per-pid context, and short-lived resource-heavy pids could aggregate into wild overestimates. 
The new chart shows **per-pid traces** plus **max/total envelopes**, and introduces a `--cpu` mode flag:

- `--cpu ps-pcpu` (default): plot ps's raw lifetime ratio untransformed. "Lossless" view; every point on the chart is an unaltered ps reading.
- `--cpu ps-cpu-timepoint`: at plot time, derive a per-interval estimate from consecutive `(pcpu, etime)` pairs to approximate instantaneous CPU. Useful when lifetime averages inflate "current" usage.

Both modes draw:

- A solid **max-across-pids** lower bound at each timestamp.
- A dashed upper bound: `total_*` from the record (RSS, and CPU in `ps-pcpu`) or sum-across-pids of the derived values (CPU in `ps-cpu-timepoint`).

Other changes:

- Drop `pmem` from the default chart; replace with absolute `rss` on a secondary axis (`twinx`).
  Host RAM is shown in the legend when `info.json` is alongside the usage file.
  Motivated by SLURM, where `pmem = rss / host_total` makes job usage look tiny on big nodes.
- Harmonize byte humanization on base 1000 / "kB/MB/GB" suffixes between the formatter and plot paths.
- Warn at runtime when `--sample-interval < 1.0`: ps reports `etime` as integer seconds, so pcpu calculations for sub-second-young pids are unstable.
- Add `docs/resource-statistics.md` covering pcpu/rss semantics, how aggregation works, and how `con-duct plot` renders these. Linked from README.

## Issue 399

Related:
 - con/duct#399 (does **not** close it)

The 5363% peak that triggered #399 came from a single pid in the ps data, where ps's `cputime / etime` calculation went racy on a sub-second-young pid.
That's a sampler-side problem; fixing it requires a different sampling strategy and is out of scope here.
Issue #399 should stay open against that work.

## Demo

The branch was rebased to drop temporary demo commits before review.
The pre-rebase tip (bundled images, side-by-side `tmp-plot-demo.md`) is preserved on the [`plot-pcpu-fix-demos-2026-05-18` branch](https://github.com/asmacdo/duct/tree/plot-pcpu-fix-demos-2026-05-18) for reference.

[demo file on demo branch](https://github.com/asmacdo/duct/blob/plot-pcpu-fix-demos-2026-05-18/tmp-plot-demo.md)

## Caveats

- Per-pid `pcpu` in records is `max-across-samples` within a report interval, not point-in-time.
  `pcpu × etime` is therefore an upper bound on cputime, not cputime itself, so the `ps-cpu-timepoint` delta math is approximate.
  Documented.
- **Linux-only math.**
  `--cpu ps-cpu-timepoint` inverts the procps identity `pcpu = cputime/etime`, which assumes ps reports a cumulative lifetime ratio.
  macOS ps reports a 5/8-decayed EWMA, so the math is meaningless there.
  `info.json` doesn't record the host platform today; on macOS logs, `ps-cpu-timepoint` silently produces wrong numbers.
  Default `ps-pcpu` mode is unaffected.

## Release Notes

 - Change: `con-duct plot` now shows per-pid traces with max-across-pids and `totals.*` envelopes. Previously it drew only `totals.pcpu` and `totals.rss` as a single line per metric, with no per-pid breakdown.
 - Change: `con-duct plot` now shows absolute `rss` on a secondary y-axis instead of `pmem`. Motivated by SLURM, where `pmem = rss / host_total` makes job usage look tiny on big nodes.
 - Change: `con-duct plot` drops `vsz` from the secondary y-axis, since this is a rarely useful metric.
 - Add: host available RAM in the legend when `info.json` is available.
 - Add: `--cpu` argument to `con-duct plot`:
   - `--cpu ps-pcpu` (default): plot ps's raw lifetime `pcpu` per pid, untransformed. Every point on the chart is an unaltered ps reading.
   - `--cpu ps-cpu-timepoint`: at plot time, derive a per-interval estimate from consecutive `(pcpu, etime)` pairs to approximate instantaneous CPU. Sidesteps lifetime-average inflation on short-lived bursty pids.
 - Change: `con-duct run` warns when `--sample-interval` is below 1.0s. ps reports `etime` as integer seconds, so `pcpu` calculations for sub-second-young pids are unstable.
 - Change: byte humanization in `con-duct plot` is now decimal (base 1000, `kB`/`MB`/`GB`), matching the run summary. Previously the plot used base-1024 with decimal-style suffixes.
 - Docs: new [`docs/resource-statistics.md`](docs/resource-statistics.md) explains what duct's `pcpu` / `rss` / `pmem` actually measure, how aggregation works, and how `con-duct plot` renders these. Linked from README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,8 @@ options:
                         running process. Sample interval must be less than or
                         equal to report interval, and it achieves the best
                         results when sample is significantly less than the
-                        runtime of the process. (default: 1.0)
+                        runtime of the process. Values below 1.0 behave
+                        erratically. (default: 1.0)
   --report-interval REPORT_INTERVAL, --r-i REPORT_INTERVAL
                         Interval in seconds at which to report aggregated
                         data. (default: 60.0)
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@ Try it out using either `duct` or `con-duct run`:
 
 `duct` is most useful when the report-interval is less than the duration of the script.
 
+## Documentation
+
+- [Interpreting duct's resource statistics](docs/resource-statistics.md) — what `pcpu`, `rss`, and the `con-duct plot` chart actually measure, and where the numbers can mislead.
+
 ## Command Reference
 
 ### con-duct
--- a/docs/resource-statistics.md
+++ b/docs/resource-statistics.md
@@ -1,0 +1,244 @@
+# Interpreting duct's resource statistics
+
+duct records resource usage in two places:
+
+- **`usage.jsonl`**: one JSON record per report interval (default: every 60 seconds), capturing per-process and session-total stats aggregated over that window.
+- **`execution_summary`** (printed at exit and stored in `info.json`): a whole-run summary of peak and average values across the full execution.
+
+The numbers in both come from the same sampling pipeline.
+This document explains what those numbers actually measure, how `con-duct plot` renders them, and where they're trustworthy vs misleading.
+
+## How duct samples and aggregates
+
+Duct polls the monitored process tree on two independent intervals:
+
+- **`--sample-interval` (default 1.0s)**: how often duct reads per-pid stats via `ps -s <session_id>`.
+  Each read is a *sample*.
+- **`--report-interval` (default 60.0s)**: how often duct writes an aggregated record to `usage.jsonl`.
+  Each record summarizes all the samples taken during that report window.
+
+Aggregation within a report window uses **max reduction**:
+
+- For each per-pid metric, the reported value is the maximum observed across all samples of that pid in the window.
+- For each session-total metric (`total_rss`, `total_pcpu`, etc.), the reported value is the maximum observed across all samples' totals in the window.
+
+Consequences worth knowing:
+
+1. **Short spikes between samples are not recorded.**
+   A process that briefly allocates 10GB and frees it within a single sample interval is invisible to duct.
+2. **Per-pid and session-total peaks may come from different sample moments.**
+   Per-pid max-reduction and total max-reduction are independent.
+   The same record can have `stats[A].rss = X` (A's peak from one sub-sample) and `total_rss = Y` (the peak simultaneous total from another sub-sample).
+
+---
+
+## CPU — `pcpu`
+
+### What it measures
+
+On Linux, `ps -o pcpu` is computed per process as:
+
+```
+pcpu = ((utime + stime) / (now - process_start_time)) × 100
+```
+
+- `utime + stime` is **cumulative CPU time consumed by the process since it started** (kernel ticks from `/proc/[pid]/stat`).
+- `now - process_start_time` is **wall-clock time elapsed since the process started**.
+
+So `pcpu` is *the fraction of wall time the process has spent on CPU, averaged from birth until the moment of sampling*.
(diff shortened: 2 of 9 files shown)


asmacdo:
Updated requirements from @yarikoptic review:

- No per-pid color cycling, no cmd labels.
- Per metric (pcpu orange, pmem blue): faint dotted per-pid traces + bold solid "envelope" at max-of-pids per timestamp ("minimum of total use") + dashed "envelope" at sum-of-pids ("maximum of total use" — may be unbounded for now).
- Drop rss from default chart, drop top-N cap. Followup PRs can re-add rss/vsz as opt-in flags.
- 3-entry linestyle legend: solid (max envelope) / dashed (sum envelope) / dotted (per-pid).

asmacdo:
I think we should consider adding a separate scale 0-100 for pmem on the right side, since pcpu is also percent but can go many times higher for multicore, which effectively squashes pmem. The issue with giving the second axis to pmem is that it would replace rss as the second y-axis. 

We could
 - add an option to switch between pmem/rss
 - add a second plot for rss

wdyt?

yarikoptic:
previews look much better!  what about an alternative I believe I alluded to in the chat -- by default plot rss (not %mem) and state total physical ram in legend (if we know)?  Then it would the right "Memory" with abs sizes up and we would be all set -- and I feel that for memory absolute value is good.

asmacdo:
Swap rss for pmem, done. 
Adding host memory to the legend, done. (Note: if the info.json is available, we use it, otherwise we dont include. The demo includes one of each.)

Note: I fixed a bug caused by duplication. The Summary formatter used decimal humanization, the graph used binary but still used MB GB instead of MiB GiB. (Fixed by removing duplication, we now report decimal everywhere). I dont think this bugfix belongs in the PR, but is worth including during draft mode so we can review the plot changes together. I've noted removing this as a TODO in the description.

yarikoptic:
I slept on it:   I think we should make it an explicit option which would allow to switch between plotting "original" pcpu which is a cumulative (and differently across OS as on OSX), vs our "time-point-estimate".  So something like `--cpu=ps-pcpu,ps-cpu-timepoint` so if `ps-cpu-timepoint` -- we do our conversion estimate and otherwise just plotting original one (no transformation). When plotting, adjust label to become `pcpu (time-point)` to accent that it is our estimate. The rest stays exactly the same. WDYT? 

edit: and in the gallery we will do both plots side -by- side to visualize the divergences.

asmacdo:
@yarikoptic I agree. That interface seems reasonable. Im trying to imagine how this would play in the future when we may have multiple samplers, (ps regular, ps pdcpu calculated, /proc, psutil, whatever). The info.json could capture the sample type(s), so I **think** this would play nice with that, and default to the 1 on that was collected or force selection if more than sampling method one was selected.

asmacdo:
TODO https://intuit.github.io/auto/docs/generated/changelog#additional-release-notes

Issue #447, opened 2026-09-16
Remains non-trivial to get interactive backend for plotting going


yarikoptic:
As a naive user I am just at loss
```
❯ uv tool install --upgrade con-duct[all]
Resolved 23 packages in 363ms
Audited 23 packages in 0.01ms
Installed 2 executables: con-duct, duct
❯ con-duct plot logs/fmriprep-ds000030-c5/acq-2026.09.09T00.59.50+64123263_1_info.json
2026-09-16T00:11:11-0400 [INFO    ] con-duct: No .env files found
2026-09-16T00:11:12-0400 [ERROR   ] con_duct.plot: Cannot display plot: your current matplotlib backend is agg which is a not a known interactive backend.
2026-09-16T00:11:12-0400 [ERROR   ] con_duct.plot: Either set environment variable MPLBACKEND to an interactive backend or use --output to save the plot to a file instead.
2026-09-16T00:11:12-0400 [ERROR   ] con_duct.plot: For more info: https://matplotlib.org/stable/users/explain/figure/backends.html
❯ MPLBACKEND=TkAgg con-duct plot logs/fmriprep-ds000030-c5/acq-2026.09.09T00.59.50+64123263_1_info.json
2026-09-16T00:13:40-0400 [INFO    ] con-duct: No .env files found
AttributeError: module '_tkinter' has no attribute '__file__'. Did you mean: '__name__'?

The above exception was the direct cause of the following exception:

ImportError: failed to load tkinter functions

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/yoh/.local/share/uv/tools/con-duct/lib/python3.13/site-packages/matplotlib/backends/_backend_tk.py", line 27, in <module>
    from . import _tkagg
ImportError: initialization failed

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/yoh/.local/bin/con-duct", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/yoh/.local/share/uv/tools/con-duct/lib/python3.13/site-packages/con_duct/cli.py", line 557, in main
    sys.exit(execute(args))
             ~~~~~~~^^^^^^
  File "/home/yoh/.local/share/uv/tools/con-duct/lib/python3.13/site-packages/con_duct/cli.py", line 483, in execute
    result = args.func(args)
[...]

Issue #439, opened 2026-07-13
Mac-15-intel is flakey and one of them often fails scheduled tests


yarikoptic:
<img width="1850" height="338" alt="Image" src="https://github.com/user-attachments/assets/fe776455-19eb-4238-86d4-843680dea21f" />

please address to keep our mailboxes tighter

asmacdo:
<img width="1617" height="1116" alt="Image" src="https://github.com/user-attachments/assets/8cf3dc4d-7936-4acf-be08-1815db01b33b" />

PR #449, opened 2026-09-20
plot: fail fast with guidance when interactive backend fails to load

yarikoptic-gitmate:
Previously, con-duct only checked that MPLBACKEND named a known
interactive backend, not that it actually loaded. A backend that
matched by name but failed to import (e.g. TkAgg with a broken/missing
tkinter) only surfaced as a raw traceback deep inside plt.subplots(),
with no guidance (con/duct#447).

Move the interactivity check earlier, before any data loading/plot
building, and actually try to load the backend module there so a
failure is caught with an actionable message pointing at --output or
a different MPLBACKEND value, instead of an unguided traceback.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Hv454Dwv5QSLfUArTtdYyf


yarikoptic-gitmate:
`test (macos-15-intel, 3.10, py)` and `test (macos-15-intel, 3.11, py)` are failing on `test_spawn_children` (an unrelated e2e process-spawning test, already `@pytest.mark.flaky(reruns=3)`) — this matches the known pre-existing runner flakiness tracked in #439, not this PR's diff (which only touches `plot.py`/`test_plot.py`). Re-running those two jobs once to confirm.

The `oldestdeps` (Python 3.10 + matplotlib 3.5) job that was previously failing here is now green after 3bd3207.

---
_Generated by [Claude Code](https://claude.ai/code)_

yarikoptic-gitmate:
`codecov/project`: patch coverage is now 100% ("All modified and coverable lines are covered by tests") after d47fe23 removed the one genuinely dead defensive line codecov had flagged. The remaining -0.15% project-level dip is just the aggregate diluting slightly from adding new (fully-covered) code to the denominator — there's no `codecov.yml` here enforcing a stricter blocking threshold, and chasing it further would mean adding tests for pre-existing, unrelated uncovered lines elsewhere in the file rather than anything this PR's diff introduced. Considering this resolved on this PR's side.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
```diff
@@ -36,6 +37,30 @@
 
 lgr = logging.getLogger(__name__)
 
+# Order in which to probe built-in interactive backends when the user hasn't
+# pinned one via MPLBACKEND: native GUI toolkits first (matching matplotlib's
+# own auto-backend preference), then the browser-based/cairo variants
+# matplotlib's own auto-fallback doesn't consider on its own.
+_INTERACTIVE_BACKEND_PROBE_ORDER = (
```
asmacdo: Is it possible to pull this list and order from matplotlib?

asmacdo: Is it possible to pull this list and order from matplotlib?

Looked into it: the set is public (`list_builtin(BackendFilter.INTERACTIVE)`), but the order is a local literal inside `pyplot.switch_backend`. More importantly, matplotlib already runs that probe itself (`macosx, qtagg, gtk4agg, gtk3agg, tkagg, wxagg`, each via `switch_backend`, which includes the check for a display) whenever the backend is unset. So by the time we see `agg` with no `MPLBACKEND`, the first half of this list has already failed. The rest (cairo, webagg, nbagg/notebook) are ones matplotlib skips on purpose.

Re-probing with `load_backend_module` also skips that display check. With no display (e.g. ssh to a cluster), `tkagg` imports fine, then `matplotlib.use()` raises an uncaught `ImportError: Cannot load backend 'tkagg' which requires the 'tk' interactive framework, as 'headless' is currently running`. On main that case gives the clean "use --output" error.

(Checked matplotlib 3.5 too: same fallback list in `switch_backend`, and this block is gated on `backend_registry` anyway, so it only ever runs on >=3.9.)

Suggest dropping the auto-probe (b37715e) and keeping the fail-fast/guidance part, which is what #447 asked for.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
```diff
@@ -271,6 +323,94 @@ def matplotlib_plot(args: argparse.Namespace) -> int:
...
+            # matplotlib 3.9.x: use rcParams instead
+            current_backend = matplotlib.rcParams["backend"]  # type: ignore[attr-defined]
+        interactive_backends = backend_registry.list_builtin(BackendFilter.INTERACTIVE)
+
+        if current_backend not in interactive_backends:
+            if os.environ.get("MPLBACKEND"):
+                # The user pinned this backend explicitly -- report it
+                # rather than silently overriding their choice.
+                lgr.error(
+                    "Cannot display plot: MPLBACKEND is set to %s which is "
```
asmacdo: This reports the resolved backend, not what the user set. With `MPLBACKEND=tkagg` and no display, it logs `MPLBACKEND is set to agg which is not a known interactive backend`. Suggest `os.environ["MPLBACKEND"]` here.

---
_Generated by [Claude Code](https://claude.ai/code)_

test/test_plot.py
```diff
@@ -223,35 +260,95 @@ def test_matplotlib_plot_non_interactive_backend(
...
+        caplog: Any,
     ) -> None:
-        """Test that plotting without output in non-interactive backend returns error using get_backend."""
+        """When MPLBACKEND isn't set and the resolved backend is
+        non-interactive, con-duct should probe the builtin interactive
+        backends itself and use the first one that loads."""
         import matplotlib.backends
 
         if not hasattr(matplotlib.backends, "backend_registry"):
             pytest.skip("requires backend_registry (matplotlib >= 3.9)")
```
asmacdo: nit: redundant now that `@requires_backend_registry` is on this test (same at L325).

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
yarikoptic-gitmate: Confirmed — verified against matplotlib's actual source (`switch_backend`): it does a `required_interactive_framework` check *after* `load_backend_module` succeeds, raising `ImportError` when the framework can't run in the current session (e.g. "headless"). My probe only replicated the module-import half, so a candidate could report as "working" and then crash uncaught in the later unguarded `matplotlib.use()` call — reintroducing exactly the #447 failure mode, and more easily triggered.

Rather than dropping the auto-probe, I fixed the root cause in f268a4d: probe with `matplotlib.use()` itself (which runs the full `switch_backend` path including the framework check), not just `load_backend_module`. Verified `switch_backend` leaves backend state untouched before it raises, so retrying the next candidate is safe — this mirrors what matplotlib's own internal auto-fallback loop does.

I kept the feature rather than reverting it because it has real value your restricted list doesn't cover: `webagg`'s `required_interactive_framework` is `None`, so it's not excluded by the headless check and works over SSH given `tornado` is installed — confirmed live (auto-selected `webagg`, served on `127.0.0.1:8988`, no crash) in a headless sandbox with no other GUI toolkit available. Added `test_matplotlib_plot_auto_probe_skips_headless_only_failure` to cover the specific bug you found (a candidate whose module imports but whose switch raises).

Let me know if you'd still rather drop it outright — happy to go that route if you still see a reason to after this fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
yarikoptic-gitmate: Good catch, fixed in f268a4d: now uses `os.environ.get("MPLBACKEND")` for this message instead of the resolved `current_backend`.

---
_Generated by [Claude Code](https://claude.ai/code)_

test/test_plot.py
yarikoptic-gitmate: Removed in f268a4d (the test was rewritten anyway to match the `matplotlib.use()`-based probe).

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
yarikoptic: > Suggest dropping the auto-probe ([b37715e](https://github.com/con/duct/commit/b37715e905010d3a2db9a2d2e335cc4099a0a133)) and keeping the fail-fast/guidance part, which is what #447 asked for.

that was the minimal bar. Making it just work would be much better IMHO. Let's try to have auto probe and see if any side effect in real world use before dropping it

yarikoptic-gitmate: Sounds good — keeping the auto-probe as fixed in f268a4d/10dddae (all CI green). Happy to revisit if real-world use turns up a case the `matplotlib.use()`-based probe doesn't handle cleanly.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
```diff
@@ -271,6 +332,95 @@ def matplotlib_plot(args: argparse.Namespace) -> int:
...
+                    "not a known interactive backend.",
+                    mplbackend,
+                )
+                lgr.error(
+                    "Either set MPLBACKEND to an interactive backend "
+                    "(one of: %s -- whichever has its GUI toolkit installed) "
+                    "or use --output to save the plot to a file instead.",
+                    ", ".join(sorted(interactive_backends)),
+                )
```
asmacdo: Reproduced with no display: `MPLBACKEND=tkagg` logs "not a known interactive backend", but tkagg is interactive, it just can't run here. Matplotlib only exposes *why* via private API, so rather than detect it, name both likely causes (`_NO_BACKEND_HINT`, suggested at the top of the module):

```suggestion
                lgr.error(
                    "Cannot display plot: MPLBACKEND=%s could not be used here. "
                    "Unset it to let con-duct try other backends.",
                    mplbackend,
                )
                lgr.error(_NO_BACKEND_HINT)
```
(`test_matplotlib_plot_non_interactive_backend_pinned` asserts `"tkagg"` from the old backend list, so it would need updating.)

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
```diff
@@ -271,6 +332,95 @@ def matplotlib_plot(args: argparse.Namespace) -> int:
...
+                    "Cannot display plot: tried all known interactive "
+                    "matplotlib backends (%s) and none of them could be "
+                    "loaded in this environment.",
+                    ", ".join(sorted(interactive_backends)),
+                )
+                lgr.error(
+                    "Install a GUI toolkit (e.g. python3-tk for Tk, "
+                    "PyQt5/PySide6 for Qt) or use --output to save the plot "
+                    "to a file instead."
+                )
```
asmacdo: Same idea here: "Install a GUI toolkit" misleads when tk is installed and the real problem is no display.

```suggestion
                lgr.error(
                    "Cannot display plot: no interactive matplotlib backend "
                    "could be used here."
                )
                lgr.error(_NO_BACKEND_HINT)
```
(`test_matplotlib_plot_auto_probe_all_fail` asserts `"tried all known interactive"`, so it would need updating.)

---
_Generated by [Claude Code](https://claude.ai/code)_

asmacdo (approved):
Failing fast and the webagg fallback are both good improvements. IMO this is good to merge, except the error messages need to be improved (suggestions inline).

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
```diff
@@ -36,6 +37,30 @@
 
 lgr = logging.getLogger(__name__)
```
asmacdo: Shared hint for both "cannot display" error paths (used by the two suggestions below), so the wording lives in one place:

```suggestion
lgr = logging.getLogger(__name__)

_NO_BACKEND_HINT = (
    "Usually this means there is no display (e.g. ssh without X forwarding) "
    "or no GUI toolkit installed. Use --output to save to a file, install a "
    "GUI toolkit (e.g. PyQt6, or python3-tk from your OS), or install tornado "
    "to view in a browser via webagg."
)
```

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
yarikoptic-gitmate: Applied verbatim in 69069e3, and updated `test_matplotlib_plot_non_interactive_backend_pinned` accordingly.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
yarikoptic-gitmate: Applied verbatim in 69069e3, and updated `test_matplotlib_plot_auto_probe_all_fail` accordingly.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/con_duct/plot.py
yarikoptic-gitmate: Applied in 929ba04 (along with the refined "MPLBACKEND=%s ... Unset it to let con-duct try other backends" wording from your other suggestions).

---
_Generated by [Claude Code](https://claude.ai/code)_

